PR #10, opened 2021-04-23
Spherical harmonic tickers

mberz:
Implement a locator and formatter for ticks based on spherical harmonics indexed according to the ambisonics channel convention.


f-brinkmann:
@mberz closing this because it looks outdated to me. We have plot/ticker.py in the current release.

mberz:
I don't think this is implemented yet. I'll re-open.

PR #28, opened 2023-03-23
Ambisonics signal class

tluebeck:
First draft of an AmbisonicsSignal class


spharpy/ambi.py
```diff
@@ -0,0 +1,274 @@
...
+    """
+    def __init__(
+            self,
+            data,
+            sampling_rate,
+            n_max,
+            basis_type,
+            normalization,
+            channel_convention,
+            condon_shortley,
```
mberz, line 24: Maybe we can check here if this can be generalized to cover arbitrary phase conventions depending if the basis type is complex or real
```suggestion
            phase_convention='default',
```

spharpy/ambi.py
```diff
@@ -0,0 +1,274 @@
...
+
+        References
+        ----------
+        ..
+
+        """
+        # TODO: discuss if ambisonics channels always on first dimension?
+        if not data.shape[1] >= (n_max + 1) ** 2:
+            raise ValueError('Data has to few coefficients '
+                             'for N = {n_max}.')
```
mberz, line 82: I think it would make sense to go for shape[-2], as for example band filter functions add extra leading dimensions, which would require reshaping signals to keep the "SH" dimension as leading dimension.

spharpy/ambi.py
```diff
@@ -0,0 +1,274 @@
...
+        return self._channel_convention
+
+    @channel_convention.setter
+    def channel_convention(self, value):
+        if self.channel_convention is not value:
+            self._recalculate_channel_convention(value)
+
+    @property
+    def condon_shortley(self):
+        return self._condon_shortley
```
mberz, line 130: see comment above, maybe we can rename this to "phase_convention"

spharpy/ambi.py
```diff
@@ -0,0 +1,274 @@
...
+                elif normalization == "maxN":
+                    self._data[:, acn, :] *= \
+                        SphericalHarmonics.sn3d_to_maxN(acn)
+            if self._normalization == 'maxN':
+                if normalization == 'n3d':
+                    self._data[:, acn, :] *= \
+                        SphericalHarmonics.maxN_to_n3d(acn)
+                elif normalization == "sn3d":
+                    self._data[:, acn, :] *= \
+                        SphericalHarmonics.maxN_to_sn3d(acn)
```
mberz, line 161: Wouldn't it suffice here to use the functions provided as part of spharpy.spherical? Using the class seems like a bit of a detour here. 
At least acn_to_nm for example should also work for arrays, making the loop obsolete. Please check if this would be feasible for the remaining functions

spharpy/ambi.py
```diff
@@ -0,0 +1,274 @@
...
+    @channel_convention.setter
+    def channel_convention(self, value):
+        if self.channel_convention is not value:
+            self._recalculate_channel_convention(value)
+
+    @property
+    def condon_shortley(self):
+        return self._condon_shortley
+
+    def _recalculate_normalization(self, normalization):
```
mberz, line 132:
```suggestion
    def _renormalize(self, normalization):
```

mberz: Just to make it a bit shorter and more explicit that the normalization is also applied.

spharpy/ambi.py
```diff
@@ -0,0 +1,274 @@
...
+                        SphericalHarmonics.sn3d_to_maxN(acn)
+            if self._normalization == 'maxN':
+                if normalization == 'n3d':
+                    self._data[:, acn, :] *= \
+                        SphericalHarmonics.maxN_to_n3d(acn)
+                elif normalization == "sn3d":
+                    self._data[:, acn, :] *= \
+                        SphericalHarmonics.maxN_to_sn3d(acn)
+
+    def _recalculate_channel_convention(self, value):
```
mberz, line 163:
```suggestion
    def _change_channel_convention(self, value):
```
Just to make it more explicit.

spharpy/ambi.py
```diff
@@ -0,0 +1,274 @@
...
+        pass
```
mberz, line 164:
```suggestion
        raise NotImplementedError("Not implemented.")
```

:)

spharpy/ambi.py
```diff
@@ -0,0 +1,274 @@
...
+        data_nm = ambisonics_signal.time
+    else:
+        data_nm = ambisonics_signal.freq
+
+    # perform inverse transform
+    data = np.tensordot(spherical_harmonics.basis, data_nm, [1, 0])
+
+    return Signal(data, ambisonics_signal.sampling_rate,
+                  fft_norm=ambisonics_signal.fft_norm,
+                  comment=ambisonics_signal.comment)
```
mberz, line 274: Maybe this can be part of a separate PR?

spharpy/classes/ambisonics.py
```diff
@@ -0,0 +1,185 @@
...
+
+        References
+        ----------
+        ..
+
+        """
+
+        if not data.shape[-2] >= (n_max + 1) ** 2:
+            raise ValueError('Data has to few sh coefficients '
+                             'for n_max = {n_max}.')
```
f-brinkmann, line 85: Should be a single line f-string or multiple line f-string in paranthesis to avoid line breaks in the error message. Did you know this trick, which prints exactly what you want:
```suggestion
            raise ValueError(f'Data has to few sh coefficients for {n_max =}.')
```
Maybe add the highest possible n_max to the message to help the user resolve the error?

spharpy/classes/ambisonics.py
f-brinkmann, line 1: I would strongly suggest to copy getters and setters from the SH class for the shared properties: 'n_max', 'basis_type', etc.

spharpy/classes/ambisonics.py
```diff
@@ -0,0 +1,185 @@
...
+            self._normalization = normalization
+        else:
+            raise ValueError("Invalid normalization, has to be 'sn3d', "
+                             f"'n3d', or 'maxN, but is {normalization}")
+
+        if channel_convention in ['acn', 'fuma']:
+            self._channel_convention = channel_convention
+        else:
+            raise ValueError("Invalid channel convention, has to be 'acn' "
+                             f"or 'fuma', but is {channel_convention}")
```
f-brinkmann, line 106: I would copy getters and setters from the SH class once this is finished!

spharpy/classes/ambisonics.py
```diff
@@ -0,0 +1,185 @@
...
+        return self._basis_type
+
+    @property
+    def normalization(self):
+        return self._normalization
+
+    @normalization.setter
+    def normalization(self, value):
+        if self.normalization is not value:
+            self._renormalize(self, value)
```
f-brinkmann, line 127: This is nice. But we should check if *value* is valid. Probably contained in the setter from the SH class.

spharpy/classes/ambisonics.py
```diff
@@ -0,0 +1,185 @@
...
+    @channel_convention.setter
+    def channel_convention(self, value):
+        if self.channel_convention is not value:
+            self._change_channel_convention(value)
+
+    @property
+    def phase_convention(self):
+        return self._phase_convention
+
+    def _renormalize(self, normalization):
```
f-brinkmann, line 144: This is nice, but does it only work for ACN? Since this is also needed in the SH class, it might be worth to have a shared function for re-normalization

spharpy/classes/ambisonics.py
```diff
@@ -0,0 +1,185 @@
...
+
+        if self._normalization == 'maxN':
+            # convert to n3d
+            self._data[:, acn, :] /= \
+                    n3d_to_maxn(acn)
+            if normalization == "sn3d":
+                self._data[:, acn, :] *= \
+                    n3d_to_sn3d_norm(acn)
+
+    def _change_channel_convention(self, value):
```
f-brinkmann, line 175: This is nice as well, and I think we should have a shared (maybe public) function for that that the SH class uses as well.

spharpy/classes/ambisonics.py
```diff
@@ -0,0 +1,185 @@
+from pyfar import Signal
+from spharpy.spherical import n3d_to_maxn, n3d_to_sn3d_norm
+from spharpy.spherical import fuma_to_nm, acn_to_nm, nm_to_acn
+
+
+class SphericalHarmonicSignal(Signal):
+    """Class for ambisonics signals.
```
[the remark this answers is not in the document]
tluebeck: Agree, I will update the docstrings

spharpy/classes/ambisonics.py
```diff
@@ -0,0 +1,185 @@
...
+
+    Objects of this class contain data which is directly convertible between
+    time and frequency domain (equally spaced samples and frequency bins). The
+    data is always real valued in the time domain and complex valued in the
```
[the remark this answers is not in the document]
tluebeck: s.o.

spharpy/classes/ambisonics.py
```diff
@@ -0,0 +1,185 @@
...
+            bins each. Time data is converted to ``float``. Frequency is
+            converted to ``complex`` and must be provided as single
+            sided spectra, i.e., for all frequencies between 0 Hz and
+            half the sampling rate.
+        sampling_rate : double
+            Sampling rate in Hz
+        n_max : int
+            Maximum spherical harmonic order. Has to match the number of
+            coefficients, such that the number of coefficients
+            >= (n_max + 1) ** 2.
```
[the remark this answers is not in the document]
tluebeck: agree

spharpy/classes/ambisonics.py
tluebeck: But didn't we agree to only implement setters for channel convention and normalization? Anything else should be fixed in the initialization.

spharpy/classes/ambisonics.py
tluebeck: s.o.

spharpy/classes/ambisonics.py
tluebeck: yup agree.

spharpy/classes/ambisonics.py
f-brinkmann: Ah, yes True. In this case maybe it might still be worth copying the code from the SH class and put it in the __init__ (or make it a private setter)

mberz (changes requested):
The implementation is coming along well I'd say. Just the comment on the property setter and private methods.

Conceptually, I'm a bit confused why n_max was renamed to n_sh. To me it does seem like a rather arbitrary change that is now inconsistent with the remaining code-base and needs to be changed everywhere. Let's discuss this during the meeting.

spharpy/classes/ambisonics.py
```diff
@@ -0,0 +1,226 @@
...
+            raise ValueError("Invalid normalization, has to be 'sn3d', "
+                             f"'n3d', or 'maxN, but is {value}")
+        self._normalization = value
+
+    def _init_channel_convention(self, value):
+        """Set the channel order convention."""
+        if value not in ["acn", "fuma"]:
+            raise ValueError("Invalid channel convention, has to be 'acn' "
+                             f"or 'fuma', but is {value}")
+        self._channel_convention = value
```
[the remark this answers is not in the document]
tluebeck: We do not want to allow setting the properties 'basis_type' and 'n_max'. This is why we have the separate init methods, which are only called in the init

tluebeck:
closed in favor of #99

PR #30, opened 2023-07-06
Add option to normalize beamformer weights

mberz:
Add the option to normalise the beamforming weights such that they fulfil the distortionless response property.

- [x] implement normalisation function
- [x] maximum front-back ratio
- [x] max rE
- [x] update example notebook

Also fixes a phase error in max-fb-ratio when the eigenvector corresponding to the largest eigenvalue is negative.


ahms5 (commented):
dont we need to test normalize_beamforming_weights, since its public method? otherwise fine

ahms5 (approved):

PR #45, opened 2023-10-26
Add Example notebooks to Sphinx

mberz:
Integrate the example notebooks into the Spinx docs using nbsphinx. To add files located outside the docs root folder the nbsphinx-link plugin is required as well.
Since nbsphinx requires pandoc, which cannot be installed via pip, a switch to using Conda environments on readthedocs.org was also necesary. Another drawback an increase in build-time. But I'd argue it's worth it. Readthedocs.org imposes no limits on CPU-hours anyways.


ahms5 (commented):
thanks for adding, it's a great feature. I think we need to add install ``pandoc`` via conda to the Contribution where we discuss how to build the docs locally.

ahms5 (approved):

PR #47, opened 2023-10-26
Remove indexing submodule

mberz:
The indexing submodule contains only three functions, ACN indexing functions are already part of the spherical submodule. I've moved the content of the indexing submodule to the spherical submodule as well.


spharpy/spherical.py
```diff
@@ -509,3 +509,109 @@ def radiation_from_sphere(
...
+            sid_m[idx_n + 2*m-1] = n-m+1
+            sid_m[idx_n + 2*m] = -(n-m+1)
+        sid_m[idx_n + 2*n + 1] = 0
+        idx_n += 2*n+1
+
+    return sid_n, sid_m
+
+
+def sid2acn(n_max):
+    """Convert from SID channel indexing as proposed by Daniel.
```
ahms5, line 545: Who is Daniel, do we know him?

tests/test_indexing.py
```diff
@@ -1,33 +1,34 @@
 """ Tests for sh channel indexing"""
```
ahms5, line 1: should be put these tests to ``test_spherical``

ahms5 (changes requested):
thanks, just small comments

spharpy/spherical.py
mberz: I've added a reference to his thesis

tests/test_indexing.py
mberz: I've moved all tests for indexing (that is also acn related things) to `test_indexing`, to make this a bit more consistent

PR #59, opened 2023-11-16
Spherical harmonic class addition

xefonon:
Hello all, 

I've drafted the spherical harmonics class "Sphharm" with some features already included and potential for extension.
Feedback is appreciated.

For discussion

- Add option for Condon-Shortley phase convention (difference between Ahrens - Rafaely) as a property. Check also if it available for real valued sph harmonics.  (scipy includes this by default) - add not implemented in gradient
- Do we want to remove the transforms from the SH class? (done)
- Do we want to allow multideimensional coordinates or force `cdim==1` - raise error: has to be 1D (possible feature) 
- Does the caching automatically recompute the bases, gradients, inverse, etc. if for example the normalization changes 
-  --private parameter getter functions check recompute and recompute if required.
- How do we want the converters between normalizations and channel orderings to work? Should they consider AmbisonicsSignals?
- Flake8 should be checked (done)

Major:
- Check CircleCI - dev branch not up to date with requirements (i.e., ModuleNotFoundError: No module named 'spharpy._deprecation')

ToDo
- [ ] Check if weights sum to 1 with some tolerance.
- [ ] Chec summation of weights in sampling sphere (see #77)

### Comments Fabian

A few thoughts on the converters `n3d_to_maxn` and `n3d_to_sn3d_norm`
- change n_sh back to n_max and make n_max docstring more explicit

- [x] For discussion: The names are inconsistent and should be changed. Were using '2' instead of 'to' in pyfar. Do we want to be consistent across packages here and use '2' everywhere?
- [ ] A few notes on the docstrings:
  - [x] They are not converting SH but returning weights for the conversion instead. I think this is fine, since conversion os handeled in the SH class. But we should maybe start the dosctrings with ' Compute order dependent weights for converting between ...'  or something similar.
  - [ ] We could add a note that conversion of SH should usually be handeled by the SH class.
  - [ ] The docs of the return parameters should mention that this is a conversion factor, how it should be applied and how it can be applied to convert in the inverse direction.

### Meeting 08/11/24:

- Example notebook with different conventions etc.
- change n_sh back to n_max and enhance docstring (regarding n_max)


spharpy/spherical.py
```diff
@@ -1,6 +1,431 @@
...
+        channel_convention : str, optional
+            Channel ordering convention, either 'acn' or 'fuma'. The default is 'acn'.
+            (FuMa is only supported up to 3rd order)
+        inv_transform_type : str, optional
+            Inverse transform type, either 'pseudo_inverse' or 'quadrature'. The default is None.
+        normalization : str, optional
+            Normalization convention, either 'n3d', 'maxN' or 'sn3d'. The default is 'n3d'.
+            (maxN is only supported up to 3rd order)
+        """
+        self.n_max = n_max
```
f-brinkmann, line 34: I would suggest to turn n_max and all other arguments into properties having a getter and setter. In this case `self.n_max = n_max` would call the setter, which would be the best place for performing the check that is currently done in `_validate_parameters`. Because `self.n_max` could not be used as a variable anymore, we would typically store the data in `self._n_max` indicating it as private.

spharpy/spherical.py
```diff
@@ -1,6 +1,431 @@
 import numpy as np
 import scipy.special as special
 import spharpy.special as _special
+import logging as logger
+from pyfar import Coordinates
+from functools import lru_cache
+from spharpy.samplings.helpers import calculate_sampling_weights
+
+
+class SphHarm:
```
f-brinkmann, line 10: Should we be more explicit and call it `SphericalHarmonics`?

spharpy/spherical.py
```diff
@@ -1,6 +1,431 @@
...
+    def __init__(self, n_max, coords, basis_type='complex', channel_convention='acn', inv_transform_type=None,
+                 normalization='n3d'):
+        r"""
```
f-brinkmann, line 13: We typically move the docstring before the `__init__`. This way it shows up as the class documentation and we do not need to include `__init__` in the documentation, which might be looking weird for users with little Python experience.

spharpy/spherical.py
```diff
@@ -1,6 +1,431 @@
...
+        Create a spherical harmonics class for transformation between spherical harmonics and signals.
```
f-brinkmann, line 14: How about
```suggestion
        Class for spherical harmonics basis functions.

        Objects of this class can be used to obtain spherical harminics basis functions, their inverse, and ...
```

spharpy/spherical.py
```diff
@@ -1,6 +1,431 @@
...
+
+        Parameters
+        ----------
+        n_max : int,
+            Maximum spherical harmonic order
+        coords : :doc:`pf.Coordinates <pyfar:classes/pyfar.coordinates>`
```
f-brinkmann, line 20: If the inverse if 'quadrature' the coordinates must have weights. This should be mentioned. Do we need `spharpy.SamplingSphere` in this case instead of `pyfar.Coordinates`?

spharpy/spherical.py
```diff
@@ -1,6 +1,431 @@
...
+            Coordinate object with sampling points for which the basis matrix is
+            calculated
+        basis_type : str, optional
+            Type of spherical harmonic basis, either 'complex' or 'real'. The default is 'complex'.
```
f-brinkmann, line 24: We have some anyoing docstring conventions: If we refer to a parameter, we make it appear itallic. This is done by single \` quotes. If we use code we make it appear as coode e.g. \`\`'complex'\`\`

spharpy/spherical.py
```diff
@@ -1,6 +1,431 @@
...
+        channel_convention : str, optional
```
f-brinkmann, line 25: This could be a bit more detailed and explain the channel ordering for both cases

spharpy/spherical.py
f-brinkmann: Would it make sense to mention already here, the dimension of the SH matrix?

spharpy/spherical.py
```diff
@@ -1,6 +1,431 @@
...
+            Channel ordering convention, either 'acn' or 'fuma'. The default is 'acn'.
+            (FuMa is only supported up to 3rd order)
+        inv_transform_type : str, optional
```
f-brinkmann, line 28: I think some more detail would help also here, i.e., briefly mention how the inverse is gerneated in both cases.

spharpy/spherical.py
f-brinkmann: How about giving the exact definitions of all Spherical Harmonics Conventions that can be realized with the class at the beginning of the docstring?

spharpy/spherical.py
```diff
@@ -1,6 +1,431 @@
...
+        if self.basis_type == 'complex':
+            self.basis = spherical_harmonic_basis(self.n_max, self.coords,
+                                                  self.normalization, self.channel_convention)
+        elif self.basis_type == 'real':
+            self.basis = spherical_harmonic_basis_real(self.n_max, self.coords,
+                                                       self.normalization, self.channel_convention)
+        else:
+            raise ValueError("Invalid signal type, should be either 'complex' or 'real'")
+
+    def compute_basis(self):
```
f-brinkmann, line 73: Maybe call this only `basis`, since the user does not care if it is actually computed each time this is called? In any case, it should have a docstring.

spharpy/spherical.py
```diff
@@ -1,6 +1,431 @@
...
+        elif self.inv_transform_type == 'pseudo_inverse':
+            self._basis_inv_gradient_theta = np.linalg.pinv(self.basis_gradient_theta)
+            self._basis_inv_gradient_phi = np.linalg.pinv(self.basis_gradient_phi)
+        elif self.inv_transform_type == 'quadrature':
+            if weights is None:
+                weights = calculate_sampling_weights(self.coords)
+            self._basis_inv_gradient_theta = np.conj(self.basis_gradient_theta).T * (4 * np.pi * weights)
+            self._basis_inv_gradient_phi = np.conj(self.basis_gradient_phi).T * (4 * np.pi * weights)
+
+    def transform(self, signal):
```
f-brinkmann, line 172: Another general question: Should the transform functions below be part of the SH class or would they go somewhere else? I guess with the inverse defined, they could be part. But aren't there cases where radial functions are combined with the SH matrix before the transform?

spharpy/spherical.py
```diff
@@ -1,6 +1,431 @@
...
+        # Implement filtering algorithm using the basis matrix
+        # TODO: Implement filtering
+        pass
+
+    def plot_basis_functions(self, order, mode):
+        # Implement plotting function for basis functions
+        # TODO: Implement plotting function for basis functions
+        pass
+
+    def plot_reconstructed_signal(self, coefficients):
```
f-brinkmann, line 207: Does this require a dedictaed plotting function or could `pyfar.plot` be used for this?

spharpy/spherical.py
```diff
@@ -1,6 +1,431 @@
...
+        # Implement plotting function for reconstructed signal
+        # TODO: Implement plotting function for reconstructed signal
+        pass
+
+
+def to_maxN_norm(acn):
```
f-brinkmann, line 213: We could also discuss, if we want to move the functions belwo into the SH class or if they are usefull outside of it as well.

f-brinkmann (changes requested):
Thanks for the effort. I mainly left concpetual questions for discussing in the group. And I wondered if we have functions for returning the index of a specific order *n* and degree *m* inside a SH matrix? That might come in handy in some cases.

spharpy/spherical.py
ahms5: the ``spharpy.SamplingSphere`` already contains an n_max, maybe we can use this ``n_max``. If we dont use the ``spharpy.SamplingSphere`` here, then why do we need it?

spharpy/spherical.py
ahms5: agreed

spharpy/spherical.py
ahms5: agreed

spharpy/spherical.py
```diff
@@ -1,6 +1,431 @@
...
+        except Exception as e:
+            raise ValueError("Error computing basis gradient:", e) from e
+
+    @property
+    def basis_inv(self):
+        if self._basis_inv is None:
+            self.compute_inverse()
+        return self._basis_inv
+
+    def compute_inverse(self, inv_transform_type=None, weights=None):
```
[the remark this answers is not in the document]
ahms5: the weights could be taken from the ``spharpy.SamplingSphere`` object.

spharpy/spherical.py
```diff
@@ -1,6 +1,431 @@
...
+        if inv_transform_type != self.inv_transform_type:
+            self.inv_transform_type = inv_transform_type
+        if self.inv_transform_type is None:
+            ValueError("Inverse transform type not specified")
+        elif self.inv_transform_type == 'pseudo_inverse':
+            self._basis_inv_gradient_theta = np.linalg.pinv(self.basis_gradient_theta)
+            self._basis_inv_gradient_phi = np.linalg.pinv(self.basis_gradient_phi)
+        elif self.inv_transform_type == 'quadrature':
+            if weights is None:
+                weights = calculate_sampling_weights(self.coords)
```
ahms5, line 168: Maybe during setting/init the ``spharpy.SamplingSphere``, we could check if it contains ``weigths`` and ``n_max``, otherwise raise an error. Then the user need to take care of it.

spharpy/spherical.py
```diff
@@ -1,6 +1,431 @@
...
+        if self.basis_gradient_theta is None or self.basis_gradient_phi is None:
+            self.compute_basis_gradient()
+        return np.dot(self.basis_gradient_theta.T, signal), np.dot(self.basis_gradient_phi.T, signal)
+
+    def inverse_transform(self, coefficients):
+        if self.basis_inv is None:
+            self.compute_inverse()
+        return np.dot(self.basis_inv.T, coefficients)
+
+    def inverse_transform_gradient(self, coefficients):
```
ahms5, line 187: Does it make sence to replace this runctions by oberoading ``*`` or sth. like this?

ahms5 (changes requested):
Thanks for implementing, maybe we should also discuss to move this to shparpy/classes/SphericalHarmonics.py, and bring it to toplevl. see pyfar.

spharpy/spherical.py
tluebeck: I think it would be good to define n_max and the sampling scheme separately. There are some use cases where we don't want to transform at the maximum order

spharpy/spherical.py
tluebeck: yes, I would defenetly remove the the transform from the SH class, I think that is a separate method

spharpy/spherical.py
```diff
@@ -1,94 +1,455 @@
...
+        if self.basis is None:
+            self.compute_basis()
+        return np.dot(self.basis.T, signal)
+
+    def transform_gradient(self, signal):
+        if self.basis_gradient_theta is None or self.basis_gradient_phi is None:
+            self.compute_basis_gradient()
+        return np.dot(self.basis_gradient_theta.T, signal), np.dot(self.basis_gradient_phi.T, signal)
+
+    def inverse_transform(self, coefficients):
```
tluebeck, line 322: s.o.

spharpy/spherical.py
```diff
@@ -1,6 +1,431 @@
...
+        # Implement interpolation algorithm using the basis matrix
+        # TODO: Implement interpolation algorithm using the basis matrix
+        pass
+
+    def filter(self, coefficients, band_type, cutoff_order):
+        # Implement filtering algorithm using the basis matrix
+        # TODO: Implement filtering
+        pass
+
+    def plot_basis_functions(self, order, mode):
```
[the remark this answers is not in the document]
tluebeck: agree

spharpy/spherical.py
```diff
@@ -1,94 +1,455 @@
...
+    return 1 / np.sqrt(2 * n + 1)
 
-    n = n.astype(int, copy=False)
-    m = m.astype(int, copy=False)
-
-    return n, m
 
-
-def nm2acn(n, m):
+def fuma2nm(fuma):
```
tluebeck, line 415: probably make it similar to above methods "to_xxx" instead of "2xxx". So renaming to fume_to_nm would be more consistent to me.

spharpy/spherical.py
```diff
@@ -130,21 +490,180 @@ def spherical_harmonic_basis(n_max, coords):
...
+            basis[:, acn] *= to_maxN_norm(acn)
+
+    return basis
+
+
+def spherical_harmonic_basis_real(n_max, coords, normalization='n3d', channel_convention='acn'):
+    r"""
+    Calculates the real valued spherical harmonic basis matrix.
+
+    The spherical harmonic functions are fully normalized (N3D) and follow
```
tluebeck, line 519: this doctoring is not valid when the user puts sn3d as a normalization or am I wrong?

spharpy/spherical.py
```diff
@@ -1,94 +1,455 @@
...
+    # Properties
+    @property
+    def n_max(self):
+        return self._n_max
+
+    @n_max.setter
+    def n_max(self, value):
+        if value < 0:
+            raise ValueError("n_max must be a positive integer")
+        self._n_max = value
```
tluebeck, line 113: Probably force a cast to integer to avoid any float issues in the following?

spharpy/spherical.py
```diff
@@ -130,21 +490,180 @@ def spherical_harmonic_basis(n_max, coords):
...
+    n = (np.ceil(np.sqrt(acn + 1)) - 1)
+    m = acn - n ** 2 - n
+
+    n = n.astype(int, copy=False)
+    m = m.astype(int, copy=False)
+
+    return n, m
+
+
+def nm2acn(n, m):
```
tluebeck, line 627: s.o.

spharpy/spherical.py
```diff
@@ -1,94 +1,455 @@
...
+        - ``'pseudo_inverse'``: Uses the Moore-Penrose pseudo-inverse for the inverse transform.
+        - ``'quadrature'``: Uses quadrature for the inverse transform.
+        - ``None``: No inverse transform is applied.
 
+    - normalization: Defines the normalization convention. It can be 'n3d', 'maxN', or 'sn3d'.
+        - ``'n3d'``: Uses the 3D normalization (also known as Schmidt semi-normalized).
+        - ``'maxN'``: Uses the maximum norm (also known as fully normalized).
+        - ``'sn3d'``: Uses the SN3D normalization (also known as Schmidt normalized).
+    """
+    def __init__(self, coords, basis_type='complex', channel_convention='acn', inv_transform_type=None,
```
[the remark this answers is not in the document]
xefonon: agreed, will change to real :)

spharpy/spherical.py
```diff
@@ -258,7 +778,9 @@ def spherical_harmonic_basis_real(n_max, coords):
         Real valued spherical harmonic basis matrix.
 
 
-    """
+    """ # noqa: 501
+    coords = spharpy.SamplingSphere.from_pyfar(coords)
+
     n_coeff = (n_max+1)**2
```
tluebeck, line 784: spaces missing: (n_max + 1) ** 2

tluebeck (changes requested):
Looks good to me, thanks for the effort. I would only remove the transform from the SH class. Or did we decide to include here?

spharpy/spherical.py
```diff
@@ -1,94 +1,455 @@
...
+
+    @weights.setter
+    def weights(self, weights):
+        if weights is None:
+            self._weights = None
+            return
+        if len(weights) != len(self.coords.cartesian):
+            raise ValueError("The number of weights has to be equal to \
+                    the number of sampling points.")
+
```
f-brinkmann, line 128: This would be a bit more complicated if we were to allow multidimensional coordinates. In any case, I think we should use `coords.cshape` or `coords.csize` to check this. and also explicitly check the shape of weights.

spharpy/spherical.py
```diff
@@ -1,94 +1,455 @@
...
+        weights = np.asarray(weights, dtype=float)
+        norm = np.linalg.norm(weights, axis=-1)
+
+        if not np.allclose(norm, 4*np.pi):
+            weights *= 4*np.pi/norm
+
```
f-brinkmann, line 134: We could also directly applay this `weights = weights / np.sum(weights) * 4 * np.pi`. And we should mention the normalization in the docstring.

spharpy/spherical.py
```diff
@@ -1,94 +1,455 @@
...
+        self._weights = weights
+    @property
+    def coords(self):
+        return self._coords
+
+    @coords.setter
+    def coords(self, value):
+        if not isinstance(value, pf.Coordinates):
+            raise TypeError("coords must be a pyfar.Coordinates object")
```
f-brinkmann, line 143: - See above: lets not store sampling sphere and coordinates. We should handle both here
- For discussion: Shoud we force `coords.cdim == 1`? Or do we want and would everything work with multidimensional Coordinates?

spharpy/spherical.py
f-brinkmann: Should be removed, right?

spharpy/spherical.py
```diff
@@ -1,94 +1,455 @@
...
+    @basis_gradient_theta.setter
+    def basis_gradient_theta(self, value):
+        self._basis_gradient_theta = value
+
+    @basis_gradient_phi.setter
+    def basis_gradient_phi(self, value):
+        self._basis_gradient_phi = value
+
+
+def to_maxN_norm(acn):
```
f-brinkmann, line 365: I would call this `n3d_to_maxn`

spharpy/spherical.py
```diff
@@ -1,94 +1,455 @@
...
+    """
+    Converts N3D normalization to max N normalization
     Parameters
```
f-brinkmann, line 368:
```suggestion
    Convert N3D normalization to max N normalization.
```

f-brinkmann (changes requested):
Thanks for the effort, it is way better already. In addition to my comments, I marked some points for discussion in the description of the pull requetst

spharpy/spherical.py
xefonon: You mean the comment or the compute_* properties?

spharpy/spherical.py
xefonon: OK!

spharpy/spherical.py
xefonon: removing!

spharpy/spherical.py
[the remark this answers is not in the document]
xefonon: OK!

spharpy/spherical.py
```diff
@@ -1,94 +1,455 @@
...
+        Y_{nm} = N_{nm} P_{nm}(cos(\theta)) e^{im\phi}
 
-    The linear index corresponds to the Ambisonics Channel Convention [#]_.
+    where:
+    - n is the degree
+    - m is the order
+    - Pnm is the associated Legendre function
+    - Nnm is the normalization term
+    - \theta is the colatitude (angle from the positive z-axis)
+    - \phi is the azimuth (angle from the positive x-axis in the xy-plane)
```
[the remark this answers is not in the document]
xefonon: Not sure how to link pyfar Coordinates documentation using intersphinx, maybe talk about it tomorrow.

spharpy/spherical.py
```diff
@@ -1,94 +1,455 @@
...
+import pyfar as pf
+from functools import lru_cache
 
-def acn2nm(acn):
+class SphericalHarmonics:
     r"""
-    Calculate the order n and degree m from the linear coefficient index.
+    Spherical Harmonics class for computing spherical harmonic basis matrices and their gradients.
+    The spherical harmonic Ynm is given by:
+    #TODO: change unicode to latex (i.e. \theta to \theta)
```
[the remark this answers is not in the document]
xefonon: has been done

spharpy/spherical.py
```diff
@@ -1,94 +1,455 @@
...
+        self.coords = coords  # coordinates : :doc:`pf.Coordinates <pyfar:classes/pyfar.coordinates>`
+        self.basis_type = basis_type
+        self.inv_transform_type = inv_transform_type
+        self.channel_convention = channel_convention  # channel ordering convention
+        self.normalization = normalization  # gain normalization convention
+        self.basis = None
+        self.basis_gradient_theta, self.basis_gradient_phi = (None, None)
+        self._basis_inv = None
+        self._basis_inv_gradient_theta, self._basis_inv_gradient_phi = (None, None)
+    # Properties
```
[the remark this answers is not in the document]
xefonon: I will remove Sampling sphere then

spharpy/spherical.py
xefonon: turn compute_basis into a private function (and compute_gradietn, etc)

spharpy/spherical.py
```diff
@@ -1,104 +1,585 @@
...
+         self.basis_gradient_phi) = (None, None)
+        self._basis_inv = None
+        (self._basis_inv_gradient_theta,
+         self._basis_inv_gradient_phi) = (None, None)
+        self._condon_shortley = condon_shortley
+        # Store previous values for comparison
+        self._prev_n_max = self.n_max
+        self._prev_coords = self.coords
+        self._prev_basis_type = self.basis_type
+        self._prev_inverse_transform = self.inverse_transform
```
mberz, line 155: This can lead to memory explosion for large sampling grids (I.e. HRTF or directivity data interpolation via SH)
Is this really necessary?

spharpy/spherical.py
```diff
@@ -1,104 +1,585 @@
...
+                    the number of sampling points."
+            )
+
+        weights = np.asarray(weights, dtype=float)
+        norm = np.linalg.norm(weights, axis=-1)
+
+        if not np.allclose(norm, 4 * np.pi):
+            weights *= 4 * np.pi / norm
+
+        self._weights = weights
```
mberz, line 195:
```suggestion
```
I'd prefer if the weights of the coordinates are used in this class.
Otherwise this leads to confusion, which set of weights is to be used, rendering one obsolete.

spharpy/spherical.py
```diff
@@ -1,104 +1,585 @@
...
+
+    @property
+    def coords(self):
+        return self._coords
+
+    @coords.setter
+    def coords(self, value):
```
mberz, line 202: I'd prefer if this is named coordinates
```suggestion
    def coordinates(self, value):
```

spharpy/spherical.py
```diff
@@ -1,104 +1,585 @@
...
+        weights = np.asarray(weights, dtype=float)
+        norm = np.linalg.norm(weights, axis=-1)
+
+        if not np.allclose(norm, 4 * np.pi):
+            weights *= 4 * np.pi / norm
+
+        self._weights = weights
+
+    @property
+    def coords(self):
```
mberz, line 198:
```suggestion
    def coordinates(self):
```

spharpy/spherical.py
mberz: @f-brinkmann You mean converting from one to the other?

+1 for storing the format that was passed here.

spharpy/spherical.py
```diff
@@ -1,104 +1,585 @@
...
+    def basis_type(self):
+        return self._basis_type
+
+    @basis_type.setter
+    def basis_type(self, value):
+        assert value in ["complex", "real"], (
+            "Invalid basis type, currently only 'complex' and"
+            " 'real' "
+            "are supported"
+        )
```
mberz, line 220: This raises an AssertionError, which is generally not very user friendly. We typically raise ValueErrors in cases of faulty user input.

spharpy/spherical.py
```diff
@@ -1,104 +1,585 @@
...
+    def inverse_transform(self):
+        return self._inverse_transform
+
+    @inverse_transform.setter
+    def inverse_transform(self, value):
+        assert value in ["pseudo_inverse", "quadrature", None], (
+            "Invalid inverse transform type, "
+            "currently only 'pseudo_inverse' "
+            "and 'quadrature' are supported"
+        )
```
mberz, line 233: This raises an AssertionError, which is generally not very user friendly. We typically raise ValueErrors in cases of faulty user input.

spharpy/spherical.py
```diff
@@ -1,104 +1,585 @@
...
+    @property
+    def channel_convention(self):
+        return self._channel_convention
+
+    @channel_convention.setter
+    def channel_convention(self, value):
+        assert value in ["acn", "fuma"], (
+            "Invalid channel convention, currently only 'acn' "
+            "and 'fuma' are supported"
+        )
```
mberz, line 245: This raises an AssertionError, which is generally not very user friendly. We typically raise ValueErrors in cases of faulty user input.

spharpy/spherical.py
```diff
@@ -1,104 +1,585 @@
...
+    def normalization(self):
+        return self._normalization
+
+    @normalization.setter
+    def normalization(self, value):
+        assert value in ["n3d", "maxN", "sn3d"], (
+            "Invalid normalization, "
+            "currently only 'n3d', 'maxN', 'sn3d' are "
+            "supported"
+        )
```
mberz, line 258: This raises an AssertionError, which is generally not very user friendly. We typically raise ValueErrors in cases of faulty user input.

spharpy/spherical.py
```diff
@@ -1,104 +1,585 @@
...
+            )
+        elif self.basis_type == "real":
+            self.basis = spherical_harmonic_basis_real(
+                self.n_max, self.coords,
+                self.normalization, self.channel_convention
+            )
+        else:
+            raise ValueError(
+                "Invalid signal type, should be either 'complex' or 'real'"
+            )
```
mberz, line 301: Could be simplified to sth like this.

```suggestion
        if self.basis_type == "complex":
            function = spherical_harmonic_basis
        elif self.basis_type == "real":
            function = spherical_harmonic_basis_real
        else:
            raise ValueError(
                "Invalid signal type, should be either 'complex' or 'real'"
            )
        self.basis = function(
                self.n_max, self.coords,
                self.normalization, self.channel_convention)
```

spharpy/spherical.py
```diff
@@ -1,104 +1,585 @@
...
+                    self.n_max, self.coords
+                )
+                self.basis_gradient_theta, self.basis_gradient_phi = (
+                    grad_theta,
+                    grad_phi,
+                )
+            else:
+                raise ValueError(
+                    "Invalid signal type, should be either 'complex' or 'real'"
+                )
```
mberz, line 340: same as above

spharpy/spherical.py
```diff
@@ -1,104 +1,585 @@
...
+    def compute_basis_gradient(self):
+        try:
+            return self._compute_basis_gradient()
+        except Exception as e:
+            raise ValueError("Error computing basis gradient:", e) from e
+
+    @property
+    def basis_inv(self):
+        if self._basis_inv is None:
+            self.compute_inverse()
```
mberz, line 351:
```suggestion
            self._compute_inverse()
```
I think we can get rid of the public compute inverse function and use the private one here.

spharpy/spherical.py
```diff
@@ -1,104 +1,585 @@
...
+    def basis(self, value):
+        self._basis = value
+
+    @basis_gradient_theta.setter
+    def basis_gradient_theta(self, value):
+        self._basis_gradient_theta = value
+
+    @basis_gradient_phi.setter
+    def basis_gradient_phi(self, value):
+        self._basis_gradient_phi = value
```
mberz, line 443: Shouldn't these be calculated using the respective compute methods?
```suggestion
```

spharpy/spherical.py
```diff
@@ -1,104 +1,585 @@
...
+                self.compute_inverse()
+            elif compute == 'basis_inv_gradient':
+                self.compute_inverse_gradient()
+            # And update the previous property values
+            self._prev_n_max = self.n_max
+            self._prev_coords = self.coords
+            self._prev_basis_type = self.basis_type
+            self._prev_inverse_transform = self.inverse_transform
+            self._prev_channel_convention = self.channel_convention
+            self._prev_normalization = self.normalization
```
mberz, line 467: Sorry, I think I missed the point of this method. What does this do or what is the benefit of this method?

spharpy/spherical.py
```diff
@@ -1,104 +1,585 @@
...
+        elif self.inverse_transform is None:
+            ValueError("Inverse transform type not specified")
+        if self.inverse_transform == "pseudo_inverse":
+            self._basis_inv = np.linalg.pinv(self.basis)
+        elif self.inverse_transform == "quadrature":
+            if self.weights is None:
+                print(
+                    "Warning: No weights specified for quadrature transform,"
+                    "calculating weights using voronoi tessellation of sphere"
+                )
```
mberz, line 389:
```suggestion
                warnings.warn(
                    "No weights specified for quadrature transform, "
                    "calculating weights using voronoi tessellation of sphere"
                )
```
make sure to import the warnings module :)

mberz (changes requested):
I've primarily focussed on the logic now (though I have added some comments on other stuff).
1. I thinks there's a couple of redundant ways of setting the basis matrix.
There's the private and the public compute methods, as well as setters.
I'd vor for keeping only the private compute method to simply everything.
2. The weights are re-implemented although they are already available from the Coordinates or Sampling class. For users it's quite arbitrary, since you have to look at the documentation to find out which one is used.

f-brinkmann (changes requested):
Thanks again for the effort, good steps forward I think. I only found time to look at the class for now skipped the functions outside the class

spharpy/spherical.py
```diff
@@ -130,27 +620,168 @@ def spherical_harmonic_basis(n_max, coords):
...
+
+    .. math::
+
+        n = \lfloor \sqrt{\mathrm{acn} + 1} \rfloor - 1
+
+        m = \mathrm{acn} - n^2 -n
+
+    Parameters
+    ----------
+    acn : ndarray, int
```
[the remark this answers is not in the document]
f-brinkmann: Could generally be changed. There are lots of explicit ndarray mentions in the docstrings. Taking lists and converting them to arrays when required might be more convenient.

spharpy/spherical.py
```diff
@@ -179,8 +810,8 @@ def spherical_harmonic_basis_gradient(n_max, coords):
     ----------
     n_max : int
         Spherical harmonic order
-    coordinates : :doc:`pf.Coordinates <pyfar:classes/pyfar.coordinates>`
-        Coordinate object with sampling points for which the basis matrix is
+    coords : :doc:`pf.Coordinates <pyfar:classes/pyfar.coordinates>`
+        Pyfar ``Coordinate`` object with sampling points for which the basis matrix is
```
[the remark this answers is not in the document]
f-brinkmann: would also prefer `coordinates`

spharpy/spherical.py
```diff
@@ -1,104 +1,585 @@
...
-    acn : ndarray, int
-        Linear index
+    acn : int
 
+    Returns
+    -------
+    maxN : float
+        Maximum norm for spherical harmonics of order N
+    """
+    assert acn <= 15
```
f-brinkmann, line 482: Only works up to order 15. How about numerically computing weights for higher order by the ratio of first order SH: SH_maxn(order, degree=0, arbitary angle) / SH_n3d(order, degree=0, arbitary angle)?

f-brinkmann (changes requested):

spharpy/spherical.py
xefonon: @f-brinkmann The problem is that you can also calculate the spherical harmonics inverse using regularised least squares (or another way possibly) and that may be a feature for a next version.

spharpy/spherical.py
xefonon: I think I'll leave this for another version. I don't have the time right now. :(

xefonon:
Hello, thanks for the review so far. I've tried to incorporate most of the comments in the revised code. Let me know what you think, and I can fix the CircleCI tests also.

spharpy/spherical.py
```diff
@@ -1,104 +1,599 @@
...
+        weights=None,
+        condon_shortley=True
+    ):
+        # self.n_max = n_max
+        # self.weights = weights
+        # self.coordinates = coordinates
+        # self.basis_type = basis_type
+        # self.inverse_transform = inverse_transform
+        # self.channel_convention = channel_convention  # ch. ord. conv.
+        # self.normalization = normalization  # gain norm. conv.
```
f-brinkmann, line 137: I think using the setters is better, because it would check the input. Why did you opt out of this?

spharpy/spherical.py
```diff
@@ -1,104 +1,599 @@
...
+    channel_convention : str, optional
+        Channel ordering convention, either ``'acn'`` or ``'fuma'``.
+        The default is ``'acn'``.
+        (FuMa is only supported up to 3rd order)
+    inverse_transform : str, optional
+        Inverse transform type, either ``'pseudo_inverse'`` or
+        ``'quadrature'``. The default is None.
+    weights : array-like, optional
+        Sampling weights for the quadrature transform.
+        Required if `quadrature` is chosen.
```
f-brinkmann, line 114: I think we decided to take the weights from coordinates. But Maybe I remember wrongly.

spharpy/spherical.py
```diff
@@ -1,104 +1,599 @@
...
+        - ``'quadrature'``: Uses quadrature for the inverse transform.
+        - ``None``: No inverse transform is applied.
 
     Parameters
     ----------
-    acn : ndarray, int
-        Linear index
+    n_max : int
+        Maximum spherical harmonic order
+    coordinates : :doc:`pf.Coordinates <pyfar:classes/pyfar.coordinates>`
```
f-brinkmann, line 95: Could also be a spharpy sampling sphere

spharpy/spherical.py
```diff
@@ -1,104 +1,599 @@
...
+        self._prev_basis_type = self.basis_type
+        self._prev_inverse_transform = self.inverse_transform
+        self._prev_channel_convention = self.channel_convention
+        self._prev_normalization = self.normalization
+        self._prev_condon_shortley = self.condon_shortley
+
+    # Properties
+    # property for condon_shortley
+    @property
+    def condon_shortley(self):
```
f-brinkmann, line 169: All setters and getters need docstrings. Only one is used for the docs, but they could simply be identical.

spharpy/spherical.py
```diff
@@ -1,104 +1,599 @@
...
+        (self._basis_inv_gradient_theta,
+         self._basis_inv_gradient_phi) = (None, None)
+        # Store previous values for comparison
+        self._prev_n_max = self.n_max
+        self._prev_coordinates = self.coordinates
+        self._prev_basis_type = self.basis_type
+        self._prev_inverse_transform = self.inverse_transform
+        self._prev_channel_convention = self.channel_convention
+        self._prev_normalization = self.normalization
+        self._prev_condon_shortley = self.condon_shortley
```
f-brinkmann, line 164: The previous values are not needed any more I think.

spharpy/spherical.py
```diff
@@ -1,104 +1,599 @@
...
+            self._recompute_basis = True
+            self._recompute_basis_gradient = True
+            self._recompute_inverse = True
+            self._recompute_inverse_gradient = True
+        self._condon_shortley = value
+
+    @property
+    def weights(self):
+        return self._weights
+
```
f-brinkmann, line 186: If we use coordinates.weights, we could also return those and discuss not to have a setter for this.

spharpy/spherical.py
```diff
@@ -1,104 +1,599 @@
...
+            raise TypeError("condon_shortley must be a boolean value")
+        if value != self._condon_shortley:
+            self._recompute_basis = True
+            self._recompute_basis_gradient = True
+            self._recompute_inverse = True
+            self._recompute_inverse_gradient = True
+        self._condon_shortley = value
+
+    @property
+    def weights(self):
```
f-brinkmann, line 184: The docstring here should mention that the weights are normalized to a sum of 4 pi

spharpy/spherical.py
```diff
@@ -1,104 +1,599 @@
...
+
+    @property
+    def basis_gradient_phi(self):
+        if (self._basis_gradient_phi is None or
+                self._recompute_basis_gradient):
+            self._compute_basis_gradient()
+            self._recompute_basis_gradient = False
+        return self._basis_gradient_phi
+
+    # @lru_cache(maxsize=128)
```
f-brinkmann, line 335: We should remove commented code

spharpy/spherical.py
```diff
@@ -1,104 +1,599 @@
...
+            function = spherical_harmonic_basis_real
+        else:
+            raise ValueError(
+                "Invalid signal type, should be either 'complex' or 'real'"
+            )
+        self._basis = function(
+            self.n_max, self.coordinates,
+            self.normalization, self.channel_convention)
+
+    def compute_basis(self):
```
f-brinkmann, line 352: All `compute_` functions should be private and for this example a property `basis` should call the private compute function.

f-brinkmann (changes requested):
I think we should generlly discuss the docstring and beahvior of the `*_to_*` functions.

spharpy/spherical.py
```diff
@@ -1,104 +1,599 @@
...
+        self._coordinates = value
+
+    @property
+    def basis_type(self):
+        return self._basis_type
+
+    @basis_type.setter
+    def basis_type(self, value):
+        if value not in ["complex", "real"]:
+            raise ValueError("Invalid basis type, currently only "
```
tluebeck, line 249: Maybe remove the currently? Which basis do we want to implement except real or complex?

spharpy/spherical.py
```diff
@@ -1,104 +1,599 @@
...
+                "SN3D normalization or FuMa channel ordering"
+            )
+        else:
+            if self.basis_type == "complex":
+                function = spherical_harmonic_basis_gradient
+            elif self.basis_type == "real":
+                function = spherical_harmonic_basis_gradient_real
+            else:
+                raise ValueError(
+                    "Invalid signal type, should be either 'complex' or 'real'"
```
tluebeck, line 375: invalid basis type?

spharpy/spherical.py
```diff
@@ -1,104 +1,599 @@
...
+    @condon_shortley.setter
+    def condon_shortley(self, value):
+        if not isinstance(value, bool):
+            raise TypeError("condon_shortley must be a boolean value")
+        if value != self._condon_shortley:
+            self._recompute_basis = True
+            self._recompute_basis_gradient = True
+            self._recompute_inverse = True
+            self._recompute_inverse_gradient = True
+        self._condon_shortley = value
```
mberz, line 181: As far as I remember the Condon Shortley phase is only used for complex valued SHs

spharpy/spherical.py
mberz, line 164:
```suggestion
```

spharpy/spherical.py
```diff
@@ -1,104 +1,599 @@
...
+    def __init__(
+        self,
+        n_max,
+        coordinates,
+        basis_type="real",
+        normalization="n3d",
+        channel_convention="acn",
+        inverse_transform=None,
+        weights=None,
+        condon_shortley=True
```
mberz, line 129:
```suggestion
        condon_shortley=None
```
I think this is conditional wheter the SHs are real or complex valued. I'm not aware of a real valued SH implementation using the CS phase.

spharpy/spherical.py
mberz: I'm also not the biggest fan of having two different ways of defining weights. It's not very intuitive and error prone.

mberz (changes requested):
My comments are only on the SH class implementation.

They're mostly on slimming down the class and making it a bit more accessible by removing redundancy.
One thing I'm not sure is the calculation of the inverse of the gradient. I'd argue that it's too specific and should not be part of the class, but implemented as part of code which does rely on it. I wasn't sure for example if a valid orthogonal basis relying on quadrature does also results in orthogonal transforms when considering the respective gradients.

The diff of this file is still quite messy, which makes it hard to track the changes. I've tried to clean up the diff a bit in #76. 
I think it would be useful to merge #76 into this branch and re-write the commit history before merging into main once this PR is finalised.
Please turn off auto-formatting or similar when editing the file.

PR #69, opened 2024-06-27
fix matplotlib deprecation

ahms5:
- test are failing atm
- ``matplotlib.cm.COLORMAP`` is deprecated.
- remove ``plt.show()`` in plot module, i was thinking this was the reson for the failing notebook
- notebook is still failing, now idea why, maybe we just stick to older versions of matplotlib


f-brinkmann:
Maybe try the notebook locally with different versions of matplotlib and numpy? Might also be related to numpy 2.0 maybe

mberz (approved):
Thanks for fixing!

PR #70, opened 2024-06-28
force numpy<2.0

ahms5:
see #69, no idea how to fix, It can be reproduces locally but its very odd.
since we will release a new version later, we can maybe stick to numpy<2.0 here.
feel free to check for alternatives

add matplotlib deprecation cm.viridis to cmap=plt.get_cmap('viridis')


mberz:
This is not the issue.
The issue is that aspect ratio handling doe not work properly within Matplotlib in python 3.8 in combination with numpy 2.x

Specifically, this fails:
```python
ax.set_box_aspect([1, 1, 0])
```
which for some reason works on all other python versions we're testing

mberz:
Seemingly obsolete

PR #73, opened 2024-07-15
fix ruff for cookiecutter

ahms5:


tests/test_plots.py
```diff
@@ -1,17 +1,15 @@
...
+import numpy as np
+import pytest
+from spharpy import plot
+import matplotlib.pyplot as plt
 import matplotlib as mpl
 # Use Agg backend to prevent matplotlib from creating interactive plots. This
 # has to be set before the importing matplotlib.pyplot. Use switch backend in
 # case the wrong backend has already been set.
 mpl.use('agg')
-import matplotlib.pyplot as plt
```
f-brinkmann, line 6: I think @mberz  mentioned that you have to call `mpl.use` before importing pyplot to be fully effective in testing - but I never checked if that is actually the case.

f-brinkmann (approved):
Approved, except for a question regarding a single line...

PR #74, opened 2024-07-15
fix warnings doc build

ahms5:
this should fix the build of doc, but there is still an issue and i have no clue why.


hoyer-a:
In the `dolph_chebyshev_weights()`  docs there is reference, but it is not used in the docstring. Maybe that derails sphinx.
Should be at least why #73 fails

ahms5:
already fixed

PR #78, opened 2024-12-02
Nm to fuma

tluebeck:
Implement a method to convert from nm to fuma index in the same way as the fuma2nm method from `dev_spherical_harmonics_class` branch

### DIFF
--- a/spharpy/spherical.py
+++ b/spharpy/spherical.py
@@ -88,7 +88,7 @@ def nm2acn(n, m):
     return n**2 + n + m
 
 
-def nm2fuma(n, m):
+def nm_to_fuma(n, m):
     r"""
     Calculate the FuMa channel index for a given spherical harmonic order n
     and degree m, according to the FuMa (Furse-Malham)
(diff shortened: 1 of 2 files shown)


spharpy/spherical.py
```diff
@@ -88,6 +88,40 @@ def nm2acn(n, m):
     return n**2 + n + m
 
 
+def nm2fuma(n, m):
```
mberz, line 91: Didn't we decide to change the naming conventions for this kind of function to `nm_to_fuma` to make conversion functions involving the normalizations more readable?

spharpy/spherical.py
```diff
@@ -88,6 +88,40 @@ def nm2acn(n, m):
...
+    acn = nm2acn(n, m)
+
+    if acn < 0 or acn >= len(fuma_mapping):
+        raise ValueError(
+            "nm2fuma only supports up to 3rd order"
+        )
+
+    # convert to fuma
+    fuma = fuma_mapping.index(acn)
+    return fuma
```
mberz, line 122: It may be a very nice feature to make this function work for arrays like the current implementations of nm2acn and acn2nm use. 
What do you think?

mberz (changes requested):
Thanks for implementing. 
I think it may be very beneficial to allow arrays as input data.

spharpy/spherical.py
tluebeck: yes, you're right. That would be good to have

mberz (commented):
Approved with only one minor comment

PR #93, opened 2025-01-21
Sh channel convention conversion

tluebeck:


tluebeck:
I have arranged with Xenofon to take a few things from the dev_sh branch. As for example the channel conversion and normalization conversion methods. If we merge them into dev, I'll delete them from his branch.

spharpy/spherical.py
```diff
@@ -131,6 +131,54 @@ def nm_to_fuma(n, m):
...
+    r"""
+    Calculate the spherical harmonic order n and degree m for a linear
+    coefficient index, according to the FuMa (Furse-Malham)
+    Channel Ordering Convention [#]_.
+
+    FuMa = WXYZ | RSTUV | KLMNOPQ
+    ACN = WYZX | VTRSU | QOMKLNP
+
+
+    References
```
f-brinkmann, line 144: References should go at the end of the docstring.

spharpy/spherical.py
```diff
@@ -131,6 +131,54 @@ def nm_to_fuma(n, m):
...
+    ----------
+    [#]  D. Malham, "Higher order Ambisonic systems” Space in Music –
+    Music in Space (Mphil thesis).
```
f-brinkmann, line 147: I think theres an indent missing here

spharpy/spherical.py
f-brinkmann: I would suggest to also add this reference to `nm_to_fuma`

f-brinkmann (changes requested):
Thanks for the effort. I only had smal suggestions. In the docstrings, I would suggest to replace 'ndarray' with 'array like' everywhere as it would be more consistent with other pyfar code.

tluebeck:
Thanks for the review. Regarding the ndarray vs array like comment. The rest of spharpy uses ndarray. That's probably a separate pr if we want to change that everywhere in the code?

f-brinkmann:
> Thanks for the review. Regarding the ndarray vs array like comment. The rest of spharpy uses ndarray. That's probably a separate pr if we want to change that everywhere in the code?

agreed

f-brinkmann (approved):
Thanks for the effort. Are the failing docs related to this pull? If yes, they should be fixed before merging - they only fail due to a warning as far as I could see.

xefonon:
> Thanks for the effort. Are the failing docs related to this pull? If yes, they should be fixed before merging - they only fail due to a warning as far as I could see.

I'm pretty sure develop branch is failing also. Maybe @ahms5 or @mberz  can confirm?

mberz:
There are problems on the develop branch as well. Let's fix these before merging and things become harder to untangle.

ahms5 (approved):

ahms5:
replaced #94

PR #94, opened 2025-01-31
fix tests on dev

ahms5:


tluebeck:
Thanks for the fix, but this was already fixed in sh_channel_convention_conversion #93

Issue #95, opened 2025-02-02
Plot functions have no tests


mberz:
The plot functions currently have no tests. We should change that. The minimum should be to test calling the functions with plausible inputs. 

Related to #91, which was caught due to failing example code.

Issue #91, opened 2025-01-10
Contour map plot fails


mberz:
The issue with the contour plot seems to be caused by a change in the behavior of the interpolation of the data to a uniform grid.
Either within the triangulation or the interpolation based on the triangulation.

Occurred in #79 

Confirmed bug in https://github.com/matplotlib/matplotlib/issues/29917

Fix will be part of [matplotlib v3.10.2](https://github.com/matplotlib/matplotlib/milestone/101)

xefonon:
CI tests failing because of below:

https://github.com/pyfar/spharpy/blob/develop/spharpy/plot/spatial.py#L722

>>> import spharpy
        >>> import numpy as np
        >>> coords = spharpy.samplings.equal_area(n_max=0, n_points=500)
        >>> data = np.sin(2*coords.colatitude) * np.cos(2*coords.azimuth)
        >>> spharpy.plot.contour_map(coords, data)
        
 No test currently - check matplotlib guidelines

f-brinkmann:
- Fixing the error using the current plot function seems difficult. The error is deep inside Matlotlib and might be related to specific Python and Matplotlib versions. Error seems to be related to Mollweide projection. Error does not appear if projection is disabled
- We decided to use a [different contour plot function](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.contourf.html) that is more flexible and will try to make it look the same way as the old plots with default parameters

@mberz is that fine for you?

f-brinkmann:
- The way understood it, it is a Matplotlib issue. Anton will post a minimal example there. 
- We will remove the failing example in the docstring

hoyer-a:
https://github.com/matplotlib/matplotlib/issues/29917

mberz:
Thanks @hoyer-a for reporting to matplotlib, I moved the issue to the on hold section until a solution to the issue is posted.

mberz:
A PR with a fix has been opened, so there's hope we don't need to implement a fix. https://github.com/matplotlib/matplotlib/pull/29919

hoyer-a:
So, using matplotlib 3.10.3 contour_map runs without errors again. But the issue concerning missing data at the poles remains.
The example i posted in the matplotlib issue produces a correct looking plot now, so we probably have to fix it in our `interpolate_data_on_sphere` method.

# `contour_map()` using contourf (same for pcolormesh)

[...]

PR #96, opened 2025-02-03
refactor contour_map function and update test cases with image_comparison (as per matplotlib test guidelines)

xefonon:
This PR refactors (slightly) the `contour_map` function in `spatial.py` and updates the test cases in `test_plots.py` to use the `image_comparison` decorator as per Matplotlib test guidelines (https://matplotlib.org/devdocs/devel/testing.html).

Changes Made:
Refactored `contour_map` Function:

Replaced `ax.contour `with `ax.contourf` to prevent breaking and ensure similar-ish results. This solves #91.
Updated Test Cases:

Added @image_comparison decorator to relevant test functions in test_plots.py to validate the visual output against baseline images. This solves (?) #95 
Refactored imports to follow Ruff (i.e., imports first, then comment)
Affected Files:
spatial.py
test_plots.py


xefonon:
OK so apparently my version of matplotlib is not the same as circleci's, so the images compared are "not the same" I guess.

f-brinkmann:
@xefonon I see that the tests are failing because of the image comparison. We've had problems with that in pyfar as well, where the images generated by matplotlib differed depending on the operation system. https://matplotlib.org/stable/api/testing_api.html#matplotlib.testing.decorators.image_comparison allows to set a tolerance, but the errors we get in testing seem to be quite large. We have RMS values between 7 and 10. Matplotlib gives 0.06 as an RMS error due to numerical precision.

Can you maybe check the baseline to see if everything is correct. Not sure if image comparison makes sense here.

xefonon:
> @xefonon I see that the tests are failing because of the image comparison. We've had problems with that in pyfar as well, where the images generated by matplotlib differed depending on the operation system. https://matplotlib.org/stable/api/testing_api.html#matplotlib.testing.decorators.image_comparison allows to set a tolerance, but the errors we get in testing seem to be quite large. We have RMS values between 7 and 10. Matplotlib gives 0.06 as an RMS error due to numerical precision.
> 
> Can you maybe check the baseline to see if everything is correct. Not sure if image comparison makes sense here.

I mean, I created the baseline, so I guess the results can vary depending on the version of matplotlib significantly. If no image comparison then what should we do?

f-brinkmann:
> I mean, I created the baseline, so I guess the results can vary depending on the version of one's OS significantly. If no image comparison then what should we do?

In pyfar we still generate images but inspect the changed images manually. That would be fine for me here as well. @mberz do you agree or should we try something else?

mberz:
The contour plot issue seems to be a matplotlib bug https://github.com/matplotlib/matplotlib/issues/29917

xefonon:
Lol I thought the problem with `contourf` was only that at the extremes the values are not quantized and binned, I didn't realize that it also creates the issues that @hoyer-a mentioned in [matplotlib/matplotlib#29917](https://github.com/matplotlib/matplotlib/issues/29917) for mpl 3.8.

Either way, it seems this will be sidelined for a while?

hoyer-a:
The issue with the values at the extremes actually seems to be caused by pcolormesh.
This is the result of 

```
...
# ax.contourf(xi, yi, zi, levels=levels, linewidths=0.5, colors='k',
#            vmin=limits[0], vmax=limits[1], extend=extend)
cf = ax.pcolormesh(xi, yi, zi, cmap=cmap, shading='gouraud',
                   vmin=limits[0], vmax=limits[1])
...
```

image

If this is fixed, i guess the better way to draw the filled contours is using contourf (and contour for the solid contour-lines) like this:

image

PR #99, opened 2025-02-18
Dev spherical harmonic signal class

tluebeck:
Implementation of the SphericalHarmonicsSignal class. The SphericalHarmonicsSIgnal inherits from the Signal class and hold the additional properties channel_convention, basis_type, n_max, and normalization.

### DIFF
--- a/spharpy/classes/ambisonics.py
+++ b/spharpy/classes/ambisonics.py
@@ -173,7 +173,7 @@ class SphericalHarmonicSignal(Signal):
         if value not in ["n3d", "maxN", "sn3d"]:
             raise ValueError("Invalid normalization, has to be 'sn3d', "
                              f"'n3d', or 'maxN, but is {value}")
-        acn = range(0, (self.n_max + 1) ** 2)
+        acn = np.arange((self.n_max + 1) ** 2)
 
         if self.channel_convention == "fuma":
             orders, degrees = fuma_to_nm(acn)
--- a/spharpy/classes/ambisonics.py
+++ b/spharpy/classes/ambisonics.py
@@ -1,0 +1,216 @@
+from pyfar import Signal
+from spharpy.spherical import n3d_to_maxn, n3d_to_sn3d_norm
+from spharpy.spherical import fuma_to_nm, acn_to_nm, nm_to_acn
+import numpy as np
+
+
+class SphericalHarmonicSignal(Signal):
+    """Create audio object with spherical harmonics coefficients in time or
+    frequency domain.
+
+    Objects of this class contain spherical harmonics coefficients which are
+    directly convertible between time and frequency domain (equally spaced
+    samples and frequency bins), the channel conventions `acn` and `fuma`, as
+    well as the normalizations `n3d`, `sn3d`, or `maxn`. The default
+    parameters for `basis_type`, `normalization`, and `channel_convention`
+    corresponds to the AmbiX standard. The definition of the spherical
+    harmonics basis functions is based on the scipy convention which includes
+    the Condon-Shortley phase.
+
+
+        Parameters
+        ----------
+        data : ndarray, double
+            Raw data of the spherical harmonics signal in the time or
+            frequency domain. The memory layout of data is 'C'. E.g.
+            data of ``shape = (3, 4, 1024)`` has 3 channels with 4
+            spherical harmonics coefficients with 1024 samples or frequency
+            bins each. Time data is converted to ``float``. Frequency is
+            converted to ``complex`` and must be provided as single
+            sided spectra, i.e., for all frequencies between 0 Hz and
+            half the sampling rate.
+        sampling_rate : double
+            Sampling rate in Hz
+        n_max : int
+            Maximum spherical harmonic order. Has to match the number of
+            coefficients, such that the number of coefficients
+            .. math:: >= (n_max + 1) ** 2.
+        basis_type : str
+            Type of spherical harmonic basis, either ``'complex'`` or
+            ``'real'``. The default is ``'real'``.
+        normalization : str
+            Normalization convention, either ``'n3d'``, ``'maxN'`` or
+            ``'sn3d'``. The default is ``'n3d'``.
+            (maxN is only supported up to 3rd order)
+        channel_convention : str
+            Channel ordering convention, either ``'acn'`` or ``'fuma'``.
+            The default is ``'acn'``.
+            (FuMa is only supported up to 3rd order)
+        n_samples : int, optional
+            Number of samples of the time signal. Required if domain is
+            ``'freq'``. The default is ``None``, which assumes an even number
+            of samples if the data is provided in the frequency domain.
+        domain : ``'time'``, ``'freq'``, optional
+            Domain of data. The default is ``'time'``
+        fft_norm : str, optional
+            The normalization of the Discrete Fourier Transform (DFT). Can be
+            ``'none'``, ``'unitary'``, ``'amplitude'``, ``'rms'``, ``'power'``,
+            or ``'psd'``. See :py:func:`~pyfar.dsp.fft.normalization` and [#]_
+            for more information. The default is ``'none'``, which is typically
+            used for energy signals, such as impulse responses.
+        comment : str
+            A comment related to `data`. The default is ``None``.
+
+        References
+        ----------
+        .. [#] E.G. Williams, "Fourier Acoustics", (1999), Academic Press
+        .. [#] B. Rafely, "Fundamentals of Spherical Array Processing", (2015),
+               Springer-Verlag
+        .. [#] F. Zotter, M. Frank, "Ambisonics A Practical 3D Audio Theory
+               for Recording, Studio Production, Sound Reinforcement, and
+               Virtual Reality", (2019), Springer-Verlag
+
+        """
+    def __init__(
+            self,
+            data,
+            sampling_rate,
+            n_max,
+            basis_type,
+            normalization,
+            channel_convention,
+            phase_convention=None,
+            n_samples=None,
+            domain='time',
+            fft_norm='none',
+            comment="",
+            is_complex=False):
+        """
+        Create SphericalHarmonicSignal with data, and sampling rate.
+        """
+        # set n_max
+        if n_max < 0:
+            raise ValueError("n_max must be a positive integer")
+        if n_max % 1 != 0:
+            raise ValueError("n_max must be an integer value")
+        if not data.shape[-2] >= (n_max + 1) ** 2:
+            raise ValueError('Data has to few sh coefficients for '
+                             f'{n_max=}. Highest possible n_max is '
+                             f'{int(np.sqrt(data.shape[-2]))-1}')
+        self._n_max = int(n_max)
+
+        # set basis_type
+        if basis_type == 'complex' and not is_complex:
+            raise ValueError('Data are real-valued while '
+                             'spherical harmonics bases are complex-valued.')
+
+        if basis_type not in ["complex", "real"]:
+            raise ValueError("Invalid basis type, only "
+                             "'complex' and 'real' are supported")
+        self._basis_type = basis_type
+
+        # set normalization
+        if normalization not in ["n3d", "maxN", "sn3d"]:
+            raise ValueError("Invalid normalization, has to be 'sn3d', "
+                             f"'n3d', or 'maxN, but is {normalization}")
+        self._normalization = normalization
+
+        # set channel_convention
+        if channel_convention not in ["acn", "fuma"]:
+            raise ValueError("Invalid channel convention, has to be 'acn' "
+                             f"or 'fuma', but is {channel_convention}")
+        self._channel_convention = channel_convention
+
+        # set phase convention
+        if not isinstance(phase_convention, type(None)):
+            if not isinstance(phase_convention, type(str)):
+                raise TypeError("phase_convention must be a string or None")
+            elif (not phase_convention == "Condon-Shortley"):
+                raise TypeError("Only Condon-Shortley phase convention is "
+                                "implemented yet.")
+        self._phase_convention = phase_convention
+
+        Signal.__init__(self, data, sampling_rate=sampling_rate,
+                        n_samples=n_samples, domain=domain, fft_norm=fft_norm,
+                        comment=comment, is_complex=is_complex)
+
+    @property
+    def n_max(self):
+        return self.n_max
+
+    @property
+    def basis_type(self):
+        return self._basis_type
+
+    @property
+    def normalization(self):
+        return self._normalization
+
+    @normalization.setter
+    def normalization(self, value):
+        if self.normalization is not value:
+            self._renormalize(self, value)
+        self._normalization = value
+
+    @property
+    def channel_convention(self):
+        return self._channel_convention
+
+    @property
+    def phase_convention(self):
+        return self._phase_convention
+
+    @channel_convention.setter
+    def channel_convention(self, value):
+        if value not in ["acn", "fuma"]:
+            raise ValueError("Invalid channel convention, has to be 'acn' "
+                             f"or 'fuma', but is {value}")
+
+        if self.channel_convention is not value:
+            self._change_channel_convention()
+
+    def _renormalize(self, value):
+        if value not in ["n3d", "maxN", "sn3d"]:
+            raise ValueError("Invalid normalization, has to be 'sn3d', "
+                             f"'n3d', or 'maxN, but is {value}")
+        acn = range(0, (self.n_max + 1) ** 2)
+
+        if self.channel_convention == "fuma":
+            orders, degrees = fuma_to_nm(acn)
+        else:
+            orders, degrees = acn_to_nm(acn)
+
+        if self._normalization == 'n3d':
+            if value == "sn3d":
+                self._data[:, :, ...] *= \
+                    n3d_to_sn3d_norm(degrees, orders)
+            elif value == "maxN":
+                self._data[:, :, ...] *= \
+                    n3d_to_maxn(acn)
+
+        if self._normalization == 'sn3d':
+            # convert to sn3d
+            self._data[:, :, :] /= \
+                    n3d_to_sn3d_norm(degrees, orders)
+            if value == "maxN":
+                self._data[:, acn, :] *= n3d_to_maxn(acn)
+
+        if self._normalization == 'maxN':
+            # convert to n3d
+            self._data[:, acn, :] /= \
+                    n3d_to_maxn(acn)
+            if value == "sn3d":
+                self._data[:, acn, :] *= \
+                    n3d_to_sn3d_norm(acn)
+
+    def _change_channel_convention(self):
+        n_coeffs = (self.n_max + 1) ** 2
+        if self._channel_convention == 'acn':
+            n, m = acn_to_nm(n_coeffs)
+            #  idx = nm2fuma(n, m)
+            raise NotImplementedError('not implemented')
+        elif self._channel_convention == 'fuma':
+            n, m = fuma_to_nm(n_coeffs)
+            idx = nm_to_acn(n, m)
+
+        self._data = self._data[:, idx, ...]
--- a/spharpy/classes/ambisonics.py
+++ b/spharpy/classes/ambisonics.py
@@ -1,6 +1,6 @@
 from pyfar import Signal
 from spharpy.spherical import n3d_to_maxn, n3d_to_sn3d_norm
-from spharpy.spherical import fuma_to_nm, acn_to_nm, nm_to_acn
+from spharpy.spherical import fuma_to_nm, nm_to_fuma, acn_to_nm, nm_to_acn
 import numpy as np
 
 
@@ -149,7 +149,7 @@ class SphericalHarmonicSignal(Signal):
     @normalization.setter
     def normalization(self, value):
         if self.normalization is not value:
-            self._renormalize(self, value)
+            self._renormalize(value)
         self._normalization = value
 
     @property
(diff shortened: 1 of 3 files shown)


spharpy/classes/ambisonics.py
```diff
@@ -0,0 +1,217 @@
...
+            comment="",
+            is_complex=False):
+        """
+        Create SphericalHarmonicSignal with data, and sampling rate.
+        """
+        # set n_max
+        if n_max < 0:
+            raise ValueError("n_max must be a positive integer")
+        if n_max % 1 != 0:
+            raise ValueError("n_max must be an integer value")
```
ahms5, line 95:
```suggestion
        if (n_max < 0) or (n_max % 1 != 0):
            raise ValueError("n_max must be a positive integer")
```
maybe we can combine it. Alternativly we could check isinstance(n_max, int), so we force int, as mentioned in the doc

ahms5 (changes requested):
thank you! I just ask github copilot, i think it can be quite useful, at lead i wouldn#t have noticed this kind of things. I have just reviewd the doc and things related to it, I'm not tomuch into the topic.

spharpy/classes/ambisonics.py
```diff
@@ -0,0 +1,220 @@
...
+
+class SphericalHarmonicSignal(Signal):
+    """Create audio object with spherical harmonics coefficients in time or
+    frequency domain.
+
+    Objects of this class contain spherical harmonics coefficients which are
+    directly convertible between time and frequency domain (equally spaced
+    samples and frequency bins), the channel conventions `acn` and `fuma`, as
+    well as the normalizations `n3d`, `sn3d`, or `maxn`. The default
+    parameters for `basis_type`, `normalization`, and `channel_convention`
```
f-brinkmann, line 15: For SH related concepts, it might be best to directly link to the documentation of the soon to come SH class where we will have written everything down in formulas

spharpy/classes/ambisonics.py
```diff
@@ -0,0 +1,220 @@
...
+    harmonics basis functions is based on the scipy convention which includes
+    the Condon-Shortley phase.
+
+
+        Parameters
+        ----------
+        data : ndarray, double
+            Raw data of the spherical harmonics signal in the time or
+            frequency domain. The memory layout of data is 'C'. E.g.
+            data of ``shape = (3, 4, 1024)`` has 3 channels with 4
```
f-brinkmann, line 26: must the data have three dimensions or can it have any number of dimensions and only the last two must be `(..., (n_max + 1)**2, data_points)`? Should be clarified in the docstring.

spharpy/classes/ambisonics.py
```diff
@@ -0,0 +1,220 @@
...
+            bins each. Time data is converted to ``float``. Frequency is
+            converted to ``complex`` and must be provided as single
+            sided spectra, i.e., for all frequencies between 0 Hz and
+            half the sampling rate.
+        sampling_rate : double
+            Sampling rate in Hz
+        n_max : int
+            Maximum spherical harmonic order. Has to match the number of
+            coefficients, such that the number of coefficients
+            .. math:: >= (n_max + 1) ** 2.
```
f-brinkmann, line 37: Does this need to be set by the user or could this also be a simple property/getter? Since we are storing SH data, must it not always be `number_of_coefficients = (n_max + 1)**2`?

spharpy/classes/ambisonics.py
```diff
@@ -0,0 +1,220 @@
...
+        Create SphericalHarmonicSignal with data, and sampling rate.
+        """
+        # set n_max
+        if (n_max < 0) or (n_max % 1 != 0):
+            raise ValueError("n_max must be a positive integer")
+        if not data.shape[-2] >= (n_max + 1) ** 2:
+            raise ValueError('Data has to few sh coefficients for '
+                             f'{n_max=}. Highest possible n_max is '
+                             f'{int(np.sqrt(data.shape[-2]))-1}')
+        self._n_max = int(n_max)
```
f-brinkmann, line 102: see comment above. I would suggest

```suggestion
        self._n_max = int(np.round(np.sqrt(data.shape[-2]) - 1))
```

and add a check, if the number of coefficients makes sense, maybe by checking `(self._n_max + 1)**2 == data.shape[-2]`. This would require first calling `Signal.__init__` to make sure data is valid (numeric) and is already a numpy array.

spharpy/classes/ambisonics.py
```diff
@@ -0,0 +1,220 @@
...
+            n_max,
+            basis_type,
+            normalization,
+            channel_convention,
+            condon_shortley='auto',
+            n_samples=None,
+            domain='time',
+            fft_norm='none',
+            comment="",
+            is_complex=False):
```
f-brinkmann, line 91: This is missing an entry in the docstring. But I have a more general question regarding this below. So don't write the docstring right ahead...

spharpy/classes/ambisonics.py
```diff
@@ -0,0 +1,220 @@
...
+        if not data.shape[-2] >= (n_max + 1) ** 2:
+            raise ValueError('Data has to few sh coefficients for '
+                             f'{n_max=}. Highest possible n_max is '
+                             f'{int(np.sqrt(data.shape[-2]))-1}')
+        self._n_max = int(n_max)
+
+        # set basis_type
+        if basis_type == 'complex' and not is_complex:
+            raise ValueError('Data are real-valued while '
+                             'spherical harmonics bases are complex-valued.')
```
f-brinkmann, line 107: Not sure if we should handle it this way. If the time data before the SH transform is real-valued, the time data after the SH transform will be real values as well. If the time data before the SH-transform ist complex valued, the time data after the SH transfrom will be complex-valued even if using a real-valued SH basis. I would suggest to not check this and use the definition of `is_complex` from the Sigbal class. Most likely there will never be the case of a complex SH Signal in this sense, but I don't see any harm in keeping the possibility, if we clearly distinguish between complex valued signals and real- and complex valued SH-bases.

spharpy/classes/ambisonics.py
```diff
@@ -0,0 +1,220 @@
...
+            Type of spherical harmonic basis, either ``'complex'`` or
+            ``'real'``. The default is ``'real'``.
+        normalization : str
+            Normalization convention, either ``'n3d'``, ``'maxN'`` or
+            ``'sn3d'``. The default is ``'n3d'``.
+            (maxN is only supported up to 3rd order)
+        channel_convention : str
+            Channel ordering convention, either ``'acn'`` or ``'fuma'``.
+            The default is ``'acn'``.
+            (FuMa is only supported up to 3rd order)
```
f-brinkmann, line 48: I think this is currently not checked in init.

spharpy/classes/ambisonics.py
```diff
@@ -0,0 +1,220 @@
...
+            Maximum spherical harmonic order. Has to match the number of
+            coefficients, such that the number of coefficients
+            .. math:: >= (n_max + 1) ** 2.
+        basis_type : str
+            Type of spherical harmonic basis, either ``'complex'`` or
+            ``'real'``. The default is ``'real'``.
+        normalization : str
+            Normalization convention, either ``'n3d'``, ``'maxN'`` or
+            ``'sn3d'``. The default is ``'n3d'``.
+            (maxN is only supported up to 3rd order)
```
f-brinkmann, line 44: I think this is currently not checked in init

spharpy/classes/ambisonics.py
```diff
@@ -0,0 +1,220 @@
...
+        return self._normalization
+
+    @normalization.setter
+    def normalization(self, value):
+        if self.normalization is not value:
+            self._renormalize(value)
+
+    @property
+    def channel_convention(self):
+        return self._channel_convention
```
f-brinkmann, line 159: Would move this done so getter and setter are next to each other in code.

spharpy/classes/ambisonics.py
```diff
@@ -0,0 +1,220 @@
...
+    @channel_convention.setter
+    def channel_convention(self, value):
+        if value not in ["acn", "fuma"]:
+            raise ValueError("Invalid channel convention, has to be 'acn' "
+                             f"or 'fuma', but is {value}")
+
+        if self.channel_convention is not value:
+            self._change_channel_convention()
+
+    def _renormalize(self, value):
```
f-brinkmann, line 174: This will also be something that needs to be done in the SH class. We should think about the best way to share this code. Probably by making it a function.

spharpy/classes/ambisonics.py
```diff
@@ -0,0 +1,220 @@
...
+        if self._normalization == 'maxN':
+            # convert to n3d
+            self._data[:, :, ...] /= \
+                    n3d_to_maxn(acn)[np.newaxis].T
+            if value == "sn3d":
+                self._data[:, :, ...] *= \
+                    n3d_to_sn3d_norm(orders)[np.newaxis].T
+        self._normalization = value
+
+    def _change_channel_convention(self):
```
f-brinkmann, line 209: See above. Will be needed for the SH class as well.

f-brinkmann (changes requested):
Thanks for the efort. I mostly had comments on the structure and the docstrings. Mind the comments regarding code that could also be used by the SH class.

I noted that you did not import the new class in the root `__init__.py` and that its not yet included in the docs. Is it still open for discussion where it should be imported? I would suggest to make it available as `spharpy.SphericalHarmonicSignal` in analogy to `pyfar.Signal` and add a new 'Classes' section to the docs also in analogy to pyfar.

mberz:
The docs are not updated, is that intended?

spharpy/classes/audio.py
```diff
@@ -0,0 +1,232 @@
...
+        References
+        ----------
+        .. [#] E.G. Williams, "Fourier Acoustics", (1999), Academic Press
+        .. [#] B. Rafely, "Fundamentals of Spherical Array Processing", (2015),
+               Springer-Verlag
+        .. [#] F. Zotter, M. Frank, "Ambisonics A Practical 3D Audio Theory
+               for Recording, Studio Production, Sound Reinforcement, and
+               Virtual Reality", (2019), Springer-Verlag
+
+        """
```
mberz, line 81: Why the additional indentation?

spharpy/classes/audio.py
```diff
@@ -0,0 +1,232 @@
...
+class SphericalHarmonicSignal(Signal):
+    """Create audio object with spherical harmonics coefficients in time or
+    frequency domain.
+
+    Objects of this class contain spherical harmonics coefficients which are
+    directly convertible between time and frequency domain (equally spaced
+    samples and frequency bins), the channel conventions `acn` and `fuma`, as
+    well as the normalizations `n3d`, `sn3d`, or `maxn`. The default
+    parameters for `basis_type`, `normalization`, and `channel_convention`
+    corresponds to the AmbiX standard. The definition of the spherical
```
mberz, line 16: Could you add some references for users to read up on these things?

spharpy/classes/audio.py
```diff
@@ -0,0 +1,232 @@
...
+    the Condon-Shortley phase.
+
+
+        Parameters
+        ----------
+        data : ndarray, double
+            Raw data of the spherical harmonics signal in the time or
+            frequency domain. The memory layout of data is 'C'. E.g.
+            data of ``shape = (3, 4, 1024)`` has 3 channels with 4
+            spherical harmonics coefficients with 1024 samples or frequency
```
mberz, line 27:
```suggestion
            spherical harmonic coefficients with 1024 samples or frequency
```

spharpy/classes/audio.py
```diff
@@ -0,0 +1,232 @@
...
+        channel_convention : str
+            Channel ordering convention, either ``'acn'`` or ``'fuma'``.
+            The default is ``'acn'``.
+            (FuMa is only supported up to 3rd order)
+        condon_shortley : bool or str, optional
+            Whether to include the Condon-Shortley phase term. If ``True`` or
+            ``'auto'``, Condon-Shortley is included, if ``False`` it is not
+            included. The default is ``'auto'``.
+        n_samples : int, optional
+            Number of samples of the time signal. Required if domain is
```
mberz, line 54:
```suggestion
            Number of time domain samples. Required if domain is
```

spharpy/classes/audio.py
mberz: I just found them below, so maybe just cite them in the corresponding places.

spharpy/classes/audio.py
```diff
@@ -0,0 +1,232 @@
...
+        elif condon_shortley == 'auto' and basis_type == 'real':
+            self._condon_shortley = False
+
+        Signal.__init__(self, data, sampling_rate=sampling_rate,
+                        n_samples=n_samples, domain=domain, fft_norm=fft_norm,
+                        comment=comment, is_complex=is_complex)
+
+    @property
+    def n_max(self):
+        """the maximum Spherical Harmonics order."""
```
mberz, line 146:
```suggestion
        """The maximum spherical harmonic order."""
```
Please apply to all below. Sentences start with a capital letter. nouns are lower-case.

spharpy/classes/audio.py
```diff
@@ -0,0 +1,232 @@
...
+        return self._n_max
+
+    @property
+    def basis_type(self):
+        """the type of the Spherical Harmonics bases (complex or real)."""
```
mberz, line 151:
```suggestion
        """The type of the spherical harmonic basis (``complex`` or ``real``)."""
```

spharpy/classes/audio.py
```diff
@@ -0,0 +1,232 @@
...
+            self._change_channel_convention()
+
+    def _renormalize(self, value):
+        if value not in ["n3d", "maxN", "sn3d"]:
+            raise ValueError("Invalid normalization, has to be 'sn3d', "
+                             f"'n3d', or 'maxN', but is {value}")
+        acn = np.arange((self.n_max + 1) ** 2)
+
+        if self.channel_convention == "fuma":
+            orders, _ = fuma_to_nm(acn)
```
mberz, line 193:
```suggestion
            orders = fuma_to_nm(acn)[0]
```
Writing to _ is not really advised :)

But I guess these comments should now go to the new branch where these methods are implemented as functions

spharpy/classes/audio.py
```diff
@@ -0,0 +1,232 @@
...
+        else:
+            orders, _ = acn_to_nm(acn)
```
mberz, line 195:
```suggestion
            orders = acn_to_nm(acn)[0]
```

mberz (changes requested):
Primarily docstring stuff. Please revise grammar and spelling :)

spharpy/classes/audio.py
```diff
@@ -0,0 +1,191 @@
...
+
+class SphericalHarmonicSignal(Signal):
+    """Create audio object with spherical harmonics coefficients in time or
+    frequency domain.
+
+    Objects of this class contain spherical harmonics coefficients which are
+    directly convertible between time and frequency domain (equally spaced
+    samples and frequency bins), the channel conventions `acn` and `fuma`, as
+    well as the normalizations `n3d`, `sn3d`, or `maxn`. The default
+    parameters for `basis_type`, `normalization`, and `channel_convention`
```
ahms5, line 16:
```suggestion
    well as the normalizations ``'n3d'``, ``'sn3d'``, or ``'maxn'``. The default
    parameters for ``basis_type``, ``normalization``, and ``channel_convention``
```

spharpy/classes/audio.py
```diff
@@ -0,0 +1,191 @@
...
+        bins each. Time data is converted to ``float``. Frequency is
+        converted to ``complex`` and must be provided as single
+        sided spectra, i.e., for all frequencies between 0 Hz and
+        half the sampling rate.
+    sampling_rate : double
+        Sampling rate in Hz
+    n_max : int
+        Maximum spherical harmonic order. Has to match the number of
+        coefficients, such that the number of coefficients
+        .. math:: >= (n_max + 1) ** 2.
```
ahms5, line 38:
```suggestion
        :math:`>= (n_{max} + 1) ^ 2`.
```

spharpy/classes/audio.py
```diff
@@ -0,0 +1,191 @@
...
+                        comment=comment, is_complex=is_complex)
+
+    @property
+    def n_max(self):
+        """the maximum spherical harmonic order."""
+        return self._n_max
+
+    @property
+    def basis_type(self):
+        """the type of the spherical harmonic basis (``complex`` or ``real``).
```
ahms5, line 152:
```suggestion
        """the type of the spherical harmonic basis (``'complex'`` or ``'real'``).
```

spharpy/classes/audio.py
```diff
@@ -0,0 +1,191 @@
...
+    @normalization.setter
+    def normalization(self, value):
+        """set the normalization of the spherical harmonic coefficients."""
+        if self.normalization is not value:
+            self._data = renormalize(self._data, self.channel_convention,
+                                     self.normalization, value, axis=-2)
+            self._normalization = value
+
+    @property
+    def condon_shortley(self):
```
ahms5, line 170: a docstring is missing here.

ahms5 (approved):
looks fine to me, thank you, I have just very tiny docstring comments.

and its not included in the documentation yet, but I think this has to be done later in another pr, see #116

ahms5 (approved):

spharpy/classes/audio.py
f-brinkmann: I would rather not treat this as code but simply as abbreviations
```suggestion
    samples and frequency bins), the channel conventions ACN and FUMA, as
    well as the normalizations N3D, SN3D, or MaxN. The default
    parameters for `basis_type`, `normalization`, and `channel_convention`
```

spharpy/classes/audio.py
```diff
@@ -0,0 +1,190 @@
...
+    spherical harmonics basis functions is based on the scipy convention which
+    includes the Condon-Shortley phase, [#]_, [#]_.
+
+
+    Parameters
+    ----------
+    data : ndarray, double
+        Raw data of the spherical harmonics signal in the time or
+        frequency domain. The memory layout of data is 'C'. E.g.
+        data of ``shape = (3, 4, 1024)`` has 3 channels with 4
```
f-brinkmann, line 25: must the data have three dimensions or can it have any number of dimensions and only the last two must be (..., (n_max + 1)**2, data_points)? Should be clarified in the docstring.

spharpy/classes/audio.py
```diff
@@ -0,0 +1,190 @@
...
+        bins each. Time data is converted to ``float``. Frequency is
+        converted to ``complex`` and must be provided as single
+        sided spectra, i.e., for all frequencies between 0 Hz and
+        half the sampling rate.
+    sampling_rate : double
+        Sampling rate in Hz
+    n_max : int
+        Maximum spherical harmonic order. Has to match the number of
+        coefficients, such that the number of coefficients
+        :math:`>= (n_{max} + 1) ^ 2`.
```
f-brinkmann, line 36: What is the idea of that? If it is the maximally possible SH order we could compute it from the size of data. If it is the transform order to be applied, it should be stated more clearly. It should also be mentioned that this must be smaller or equal to `np.sqrt(data.shape[-2]) - 1`.

spharpy/classes/audio.py
```diff
@@ -0,0 +1,190 @@
...
+
+        # set basis_type
+        if basis_type == 'complex' and not is_complex:
+            raise ValueError('Data are real-valued while '
+                             'spherical harmonics bases are complex-valued.')
+
+        if basis_type not in ["complex", "real"]:
+            raise ValueError("Invalid basis type, only "
+                             "'complex' and 'real' are supported")
+        self._basis_type = basis_type
```
f-brinkmann, line 115: Not sure if we should handle it this way. If the time data before the SH transform is real-valued, the time data after the SH transform will be real values as well. If the time data before the SH-transform ist complex valued, the time data after the SH transform will be complex-valued even if using a real-valued SH basis. I would suggest to not check this and use the definition of is_complex from the Signal class. Most likely there will never be the case of a complex SH Signal in this sense, but I don't see any harm in keeping the possibility, if we clearly distinguish between complex valued signals and real- and complex valued SH-bases.

spharpy/classes/audio.py
```diff
@@ -0,0 +1,190 @@
...
+        if not isinstance(condon_shortley, bool) and condon_shortley != 'auto':
+            raise ValueError(
+                "Condon_shortley has to be a bool, or 'auto'.")
+
+        if condon_shortley == 'auto' and basis_type == 'complex':
+            self._condon_shortley = True
+        elif condon_shortley == 'auto' and basis_type == 'real':
+            self._condon_shortley = False
+
+        Signal.__init__(self, data, sampling_rate=sampling_rate,
```
f-brinkmann, line 139: before or after this there should be a check that data has at least 3 dimensions. Otherwise if cant be an SH signal.

spharpy/classes/audio.py
```diff
@@ -0,0 +1,190 @@
...
+                        n_samples=n_samples, domain=domain, fft_norm=fft_norm,
+                        comment=comment, is_complex=is_complex)
+
+    @property
+    def n_max(self):
+        """the maximum spherical harmonic order."""
```
f-brinkmann, line 145: See comment in docstring. Purpose of this should be more clear

spharpy/classes/audio.py
```diff
@@ -0,0 +1,190 @@
...
+        return self._n_max
+
+    @property
+    def basis_type(self):
+        """the type of the spherical harmonic basis (``'complex'`` or ``'real'``).
+        """
```
f-brinkmann, line 151: Would be more consistent to only document the options once at the top of the class.
```suggestion
        """Get the type of the spherical harmonic basis."""
```

spharpy/classes/audio.py
```diff
@@ -0,0 +1,190 @@
...
+        return self._basis_type
+
+    @property
+    def normalization(self):
+        """the normalization of the spherical harmonic coefficients."""
```
f-brinkmann, line 156: Docstrings of setters and getters should be unversial and identical because only one of them shows up in the docs. E.g.

```suggestion
        """Get or set the normalization of the spherical harmonic coefficients."""
```

spharpy/classes/audio.py
```diff
@@ -0,0 +1,190 @@
...
+            self._normalization = value
+
+    @property
+    def condon_shortley(self):
+        """whether to include the Condon-Shortley phase term."""
+        return self._condon_shortley
+
+    @property
+    def channel_convention(self):
+        """the channel convention of the spherical harmonic coefficients."""
```
f-brinkmann, line 174: See above:
```suggestion
        """Get or set the channel convention of the spherical harmonic coefficients."""
```

spharpy/classes/audio.py
f-brinkmann: I think this is something for discussion during the weekly.

f-brinkmann (changes requested):
Thanks for the effort. Only reviewed the class but not the tests atm. Had one general question about the intention of `n_max` and I wondered if we want a function `self.channel_index(n, m)` that returns the index of one or multiple SH channels

spharpy/classes/audio.py
ahms5: agree

spharpy/classes/audio.py
```diff
@@ -0,0 +1,178 @@
...
+        ``'real'``.
+    normalization : str
+        Normalization convention, either ``'n3d'``, ``'maxN'`` or
+        ``'sn3d'``. (maxN is only supported up to 3rd order)
+    channel_convention : str
+        Channel ordering convention, either ``'acn'`` or ``'fuma'``.
+        (FuMa is only supported up to 3rd order)
+    condon_shortley : bool or str, optional
+        Whether to include the Condon-Shortley phase term. If ``True``
+        Condon-Shortley is included, if ``False`` it is not included.
```
f-brinkmann, line 43: This would be a bit more accurate for me:
```suggestion
        Flag to indicate if the Condon-Shortley phase term is included (``True``) or not (``False``).
```

spharpy/classes/audio.py
```diff
@@ -0,0 +1,178 @@
...
+        """
+        Create SphericalHarmonicSignal with data, and sampling rate.
+        """
+        # check dimensions
+        if len(data.shape) < 3:
+            raise ValueError("Invalid number of dimensions. Data should have "
+                             "at least 3 dimensions.")
+
+        # set n_max
+        self._n_max = int(np.sqrt(data.shape[-2]))-1
```
f-brinkmann, line 95: We should add a check, to make sure the number of channels is valid, e.g.
```suggestion
        n_max = np.sqrt(data.shape[-2])-1
        if n_max - int(n_max) != 0:
            raise ValueError('Bla')
        self._n_max = n_max
```

tests/test_spherical_harmonics_signal.py
f-brinkmann, line 1: I checked the coverage using `pytest --cov=. --cov-report=html` and saw that the errors are currently not tested. Can you please add that?

f-brinkmann (changes requested):
Thanks for the update. The functionality looks good to me now. I only had small comments left.

PR #82, opened 2025-01-02
Dev spherical harmonic class

xefonon:
Recreated spherical harmonics class PR (which I accidentally deleted)

PR #101, opened 2025-02-20
Add Spherical Harmonics class

mberz:
Replaces #82 

Re-created and cleaned up to make reviewing easier and keep the history tidy.

- [x] Requires #102 for the updated basis matrix functions
- [x] Requires #113

We dscussed the following during the weekly, which depends on #119:
- Sampling sphere gets a boolean property `quadrature` (to do @f-brinkmann)
- Sampling enforces that the radius is unique (to do @f-brinkmann)
- `SphericalHarmonics.inverse_transform` should be renamed to 'inverse_method'
- if SphericalHarmonicsProperty.inverse_method = 'auto':
	- For SamplingSphere: `if Coordinates.qaudrature; inverse_method = 'quadrature' else 'pseudo_inverse'`
	- For Coordinates: `inverse_method = 'auto'` cannot be used and must be specified by user
- inverse_basis will raise a warning if using a Coordinates object


tests/test_sh_class.py
```diff
@@ -0,0 +1,64 @@
+"""
```
tluebeck, line 1: shouldn't we also test the setter? phase_convention, n_max, weights, etc

spharpy/spherical.py
```diff
@@ -767,3 +768,404 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        self._n_max = None
+        self._coordinates = pf.Coordinates()
+        self._basis_type = None
+        self._weights = None
+        self._inverse_transform = None
+        self._channel_convention = None  # ch. ord. conv.
+        self._normalization = None  # gain norm. conv.
+        self._phase_convention = None
+
+        self.n_max = n_max
```
tluebeck, line 903: we need a check that when a user inits a signal with n_max > 3, fuma and maxN is not allowed. I think that is not caught anywhere yet. Probably the best place are the setters?

tluebeck (changes requested):
Thanks for the effort. Looks really good to me, I just have a few minor things.

spharpy/spherical.py
```diff
@@ -767,3 +768,404 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+         for the inverse transform.
+        - ``'quadrature'``: Uses quadrature for the inverse transform.
+        - ``None``: No inverse transform is applied.
+
+    Parameters
+    ----------
+    n_max : int
+        Maximum spherical harmonic order
+    coordinates : :doc:`pf.Coordinates <pyfar:classes/pyfar.coordinates>` or
+        `sp.SamplingSphere <spharpy:classes/spharpy.samplings.coordinates>`
```
ahms5, line 861:
```suggestion
    coordinates : :py:class:`~pyfar.classes.coordinates.Coordinates ` or :py:class:`~spharpy.samplings.coordinates.SamplingSphere`
```
the ~ makes it in short form in the doc.

spharpy/spherical.py
```diff
@@ -767,3 +768,404 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+    ):
+        # hidden attributes
+        self._n_max = None
+        self._coordinates = pf.Coordinates()
+        self._basis_type = None
+        self._weights = None
+        self._inverse_transform = None
+        self._channel_convention = None  # ch. ord. conv.
+        self._normalization = None  # gain norm. conv.
+        self._phase_convention = None
```
ahms5, line 901: we could write the comments in a way that its easy to understand for other developer.

ahms5 (changes requested):
thank you! I mostly reviewed the documentation, I think sometime we still need to convert it to numpy style and I think some things will break the Documentation build as well.

spharpy/spherical.py
```diff
@@ -767,3 +768,404 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+    - $*$ denotes the complex conjugate
+    - $\delta$ is the Kronecker delta
+    - $d\omega$ is the differential solid angle
+    - The integral is over the entire sphere
+
+    The class supports the following conventions:
+
+    - basis_type: Defines the type of spherical harmonic basis.
+    It can be either 'complex' or 'real'.
+        - ``'complex'``: Uses complex-valued spherical harmonics.
```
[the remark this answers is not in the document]
xefonon: I think it's a bit too much to describe parameters. This doc was created after several reviews where we agreed to have a detailed description of the theory.

spharpy/spherical.py
```diff
@@ -767,3 +768,404 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+    ----------
+    n_max : int
+        Maximum spherical harmonic order
+    coordinates : :doc:`pf.Coordinates <pyfar:classes/pyfar.coordinates>` or
+        `sp.SamplingSphere <spharpy:classes/spharpy.samplings.coordinates>`
+        objects with sampling points for which the basis matrix is
+        calculated
+    basis_type : str, optional
+        Type of spherical harmonic basis, either ``'complex'`` or
+        ``'real'``. The default is ``'complex'``.
```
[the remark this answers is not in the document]
xefonon: Yes we had agreed this I think.

spharpy/spherical.py
```diff
@@ -767,3 +768,404 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+
+class SphericalHarmonics:
+    r"""
+    Compute spherical harmonic basis matrices, their inverses, and gradients.
+    The spherical harmonic Ynm is given by:
+    .. math::
+        Y_{nm} = N_{nm} P_{nm}(cos(\theta)) T_{nm}(\phi)
+
+    where:
+    - $n$ is the degree
```
ahms5, line 782:
```suggestion
    where:
 
    - :math:`n` is the degree
```
in this was the bulletpoints and math would be fixed

mberz:
@xefonon #102 Has been merged into develop.

spharpy/spherical.py
```diff
@@ -767,3 +768,404 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+
+    @coordinates.setter
+    def coordinates(self, value):
+        """Set the coordinates."""
+        if not isinstance(value, pf.Coordinates):
+            raise TypeError("coordinates must be a pyfar.Coordinates "
+                            "object or spharpy.SamplingSphere object")
+        if value.cdim != 1:
+            raise ValueError("Coordinates must be 1D")
+            # TODO: Allow for 2D coordinates
```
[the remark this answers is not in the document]
xefonon: Not sure what you mean with this

xefonon:
Hey peeps! I made changes to this branch. Is it ready to merge? 

@mberz @tluebeck @ahms5 @f-brinkmann

spharpy/spherical.py
```diff
@@ -776,16 +778,17 @@ def aperture_vibrating_spherical_cap(
...
     r_sphere : double, ndarray
         Radius of the sphere
     r_cap : double
-        Radius of the vibrating cap
+      harmonics are real or complex
 
     Returns
     -------
-    A : ndarray, float
+    The normalization term :math:`N_{nm}` is given by:
```
xefonon, line 791: I can see that the docstring is nonsensical, because I applied suggested changes after merging from develop, and this is the monster I created. Will fix.

ahms5: nice, we are almost there, cool, thank you!
its hard to get the right indent during review.  so feel free to fix it. The docs would raise warning, atm, you can check that builing the  doc and adding ```SPHINXOPTS="-W"``` flag. usually this tests is excecuted on CI, but since 3.8 is failing it is not.

spharpy/spherical.py
```diff
@@ -1005,3 +1007,413 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+
+
+class SphericalHarmonics:
+    r"""
+    Compute spherical harmonic basis matrices, their inverses, and gradients.
+    The spherical harmonic Ynm is given by:
+    .. math::
+        Y_{nm} = N_{nm} P_{nm}(cos(\theta)) T_{nm}(\phi)
+
+    where:
```
ahms5, line 1020:
```suggestion
    where:

```
new line is required before and after a bulletpoint list

spharpy/spherical.py
```diff
@@ -1005,3 +1007,413 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+    The term :math:`T_{nm}` is defined as:
+        - For complex-valued harmonics:
+            .. math::
+                T_{nm} = e^{im\phi}
+        - For real-valued harmonics:
+            .. math::
+                T_{nm} = \begin{cases}
+                            \cos(m\phi) & \text{if } m \ge 0 \\
+                            \sin(m\phi) & \text{if } m < 0
+                        \end{cases}
```
ahms5, line 1054:
```suggestion

    - For complex-valued harmonics:
        .. math::
            T_{nm} = e^{im\phi}
    - For real-valued harmonics:
        .. math::
            T_{nm} = \begin{cases}
                        \cos(m\phi) & \text{if } m \ge 0 \\
                        \sin(m\phi) & \text{if } m < 0
                    \end{cases}
```
i suggest to remove the intend, then it looks more pretty in the doc and in the docstring.

spharpy/spherical.py
```diff
@@ -1005,3 +1007,413 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        Obtained from the coordinates object and are
+        normalized to a sum of 4*pi."""
+        if self._weights is not None:
+            return self._weights
+        elif hasattr(self.coordinates, 'weights') and self.coordinates.weights is not None:
+            return self.coordinates.weights
+        else:
+            # Calculate weights if not already set
+            self._weights = calculate_sampling_weights(self.coordinates)
+            return self._weights
```
f-brinkmann, line 1189: Is this the final version? As far as I remember we decided to always take the weights from the coordinates object. Silently computing them using `calculate_sampling_weights` might lead to unexpected behavior.

spharpy/spherical.py
```diff
@@ -1005,3 +1007,413 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+                            \cos(m\phi) & \text{if } m \ge 0 \\
+                            \sin(m\phi) & \text{if } m < 0
+                        \end{cases}
+
+    The spherical harmonics are orthogonal on the unit sphere, i.e.,
+
+    .. math::
+        \int_{sphere} Y_{nm} Y_{n'm'}* d\omega = \delta_{nn'} \delta_{mm'}
+
+    where:
```
ahms5, line 1061:
```suggestion
    where:

```
see above

spharpy/spherical.py
```diff
@@ -1005,3 +1007,413 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        elif hasattr(self.coordinates, 'weights') and self.coordinates.weights is not None:
+            return self.coordinates.weights
+        else:
+            # Calculate weights if not already set
+            self._weights = calculate_sampling_weights(self.coordinates)
+            return self._weights
+
+    @weights.setter
+    def weights(self, value):
+        self._weights = value
```
f-brinkmann, line 1193: If the above is true, we should discuss is we want a separate setter for the weights, or if they have to be set through a Coordinates object

spharpy/spherical.py
```diff
@@ -1005,3 +1007,413 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+
+    @n_max.setter
+    def n_max(self, value):
+        """Set the spherical harmonic order."""
+        if value < 0:
+            raise ValueError("n_max must be a positive integer")
+        if value % 1 != 0:
+            raise ValueError("n_max must be an integer value")
+        if hasattr(self, 'channel_convention') and self.channel_convention == "fuma" and value > 3:
+            raise ValueError("n_max cannot be greater than 3 with 'fuma' channel convention")
```
f-brinkmann, line 1208: Both lines are too long and I assume the check `hasattr(self, 'channel_convention')` is not needed, or is it?

spharpy/spherical.py
```diff
@@ -1005,3 +1007,413 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+    The class supports the following conventions:
+
+    - normalization: Defines the normalization convention.
+     It can be 'n3d', 'maxN', or 'sn3d'.
+        - ``'n3d'``: Uses the 3D normalization
+        (also known as Schmidt semi-normalized).
+        - ``'maxN'``: Uses the maximum norm
+        (also known as fully normalized).
+        - ``'sn3d'``: Uses the SN3D normalization
+        (also known as Schmidt normalized).
```
ahms5, line 1076:
```suggestion
    - basis_type: Defines the type of spherical harmonic basis.

      - ``'complex'``: Uses complex-valued spherical harmonics.
      - ``'real'``: Uses real-valued spherical harmonics.

    - ``'normalization'``: Defines the normalization convention:

      - ``'n3d'``: Uses the 3D normalization
        (also known as Schmidt semi-normalized).
      - ``'maxN'``: Uses the maximum norm
        (also known as fully normalized).
      - ``'sn3d'``: Uses the SN3D normalization
        (also known as Schmidt normalized).
```
this fixes the intend, was it intended to remove ``'basis_type'``?

spharpy/spherical.py
```diff
@@ -1005,3 +1007,413 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+    @n_max.setter
+    def n_max(self, value):
+        """Set the spherical harmonic order."""
+        if value < 0:
+            raise ValueError("n_max must be a positive integer")
+        if value % 1 != 0:
+            raise ValueError("n_max must be an integer value")
+        if hasattr(self, 'channel_convention') and self.channel_convention == "fuma" and value > 3:
+            raise ValueError("n_max cannot be greater than 3 with 'fuma' channel convention")
+        if value != self._n_max:
```
f-brinkmann, line 1209: Maybe more robust
```suggestion
        if int(value) != self._n_max:
```

spharpy/spherical.py
```diff
@@ -1005,3 +1007,413 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        - ``'maxN'``: Uses the maximum norm
+        (also known as fully normalized).
+        - ``'sn3d'``: Uses the SN3D normalization
+        (also known as Schmidt normalized).
+
+    - channel_convention: Defines the channel ordering convention.
+    It can be either 'acn' or 'fuma'.
+        - ``'acn'``: Follows the Ambisonic Channel Number (ACN) convention.
+        - ``'fuma'``: Follows the Furse-Malham (FuMa) convention.
+        (FuMa is only supported up to 3rd order)
```
ahms5, line 1082:
```suggestion
    - ``'channel_convention'``: Defines the channel ordering convention.
      
      - ``'acn'``: Follows the Ambisonic Channel Number (ACN) convention.
      - ``'fuma'``: Follows the Furse-Malham (FuMa) convention.
        (FuMa is only supported up to 3rd order)
```
i think the sentence in line 1079 is not required, its kind of clear from the list, if we want to keep it we need to add two spaces there, otherwise the format of the list will be broken.

spharpy/spherical.py
```diff
@@ -1005,3 +1007,413 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        if value % 1 != 0:
+            raise ValueError("n_max must be an integer value")
+        if hasattr(self, 'channel_convention') and self.channel_convention == "fuma" and value > 3:
+            raise ValueError("n_max cannot be greater than 3 with 'fuma' channel convention")
+        if value != self._n_max:
+            self._recompute_basis = True
+            self._recompute_basis_gradient = True
+            self._recompute_inverse = True
+            self._recompute_inverse_gradient = True
+        self._n_max = int(value)  # Cast to int for safety
```
f-brinkmann, line 1214: Could be moved into the if case.
```suggestion
            self._n_max = int(value)  # Cast to int for safety
```

spharpy/spherical.py
```diff
@@ -1005,3 +1007,413 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        - ``'acn'``: Follows the Ambisonic Channel Number (ACN) convention.
+        - ``'fuma'``: Follows the Furse-Malham (FuMa) convention.
+        (FuMa is only supported up to 3rd order)
+
+    - inverse_transform: Defines the type of inverse transform.
+     It can be 'pseudo_inverse', 'quadrature', or None.
+        - ``'pseudo_inverse'``: Uses the Moore-Penrose pseudo-inverse
+         for the inverse transform.
+        - ``'quadrature'``: Uses quadrature for the inverse transform.
+        - ``None``: No inverse transform is applied.
```
ahms5, line 1089:
```suggestion
    - ``'inverse_transform'``: Defines the type of inverse transform.

      - ``'pseudo_inverse'``: Uses the Moore-Penrose pseudo-inverse
        for the inverse transform.
      - ``'quadrature'``: Uses quadrature for the inverse transform.
      - ``None``: No inverse transform is applied.
```

spharpy/spherical.py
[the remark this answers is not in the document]
f-brinkmann: I think he ment is the TODO still valid. For now I would suggest to remove the TODO and extend for 2D at a later point, if required. I don't see a use for that. Do you have examples, where this would be needed?

spharpy/spherical.py
```diff
@@ -1005,3 +1007,413 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+                            "object or spharpy.SamplingSphere object")
+        if value.cdim != 1:
+            raise ValueError("Coordinates must be 1D")
+            # TODO: Allow for 2D coordinates
+        if value != self._coordinates:
+            self._recompute_basis = True
+            self._recompute_basis_gradient = True
+            self._recompute_inverse = True
+            self._recompute_inverse_gradient = True
+        self._coordinates = value
```
f-brinkmann, line 1235: Move inside if case
```suggestion
            self._coordinates = value
```

spharpy/spherical.py
```diff
@@ -1005,3 +1007,413 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        """
+        if value not in ["complex", "real"]:
+            raise ValueError("Invalid basis type, only "
+                             "'complex' and 'real' are supported")
+        if value != self._basis_type:
+            self._recompute_basis = True
+            self._recompute_basis_gradient = True
+            self._recompute_inverse = True
+            self._recompute_inverse_gradient = True
+        self._basis_type = value
```
f-brinkmann, line 1256: Move inside the if case

spharpy/spherical.py
```diff
@@ -1005,3 +1007,413 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+         for the inverse transform.
+        - ``'quadrature'``: Uses quadrature for the inverse transform.
+        - ``None``: No inverse transform is applied.
+
+    Parameters
+    ----------
+    n_max : int
+        Maximum spherical harmonic order
+    coordinates : :doc:`pf.Coordinates <pyfar:classes/pyfar.coordinates>` or
+        `sp.SamplingSphere <spharpy:classes/spharpy.samplings.coordinates>`
```
ahms5, line 1096:
```suggestion
    coordinates : :py:class:`pyfar.Coordinates`, spharpy.SamplingSphere
```
this shpuld work for pyfar, and for spharpy the link does not exist, so i think this is not part of the pr

spharpy/spherical.py
```diff
@@ -1005,3 +1007,413 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        """Set the inverse transform type."""
+        if value not in ["pseudo_inverse", "quadrature", None]:
+            raise ValueError("Invalid inverse transform type. " \
+            "Allowed values are 'pseudo_inverse', 'quadrature', or None.")        
+        if value != self._inverse_transform:
+            self._recompute_basis = True
+            self._recompute_basis_gradient = True
+            self._recompute_inverse = True
+            self._recompute_inverse_gradient = True
+        self._inverse_transform = value
```
f-brinkmann, line 1274: Move inside the if case

spharpy/spherical.py
```diff
@@ -1005,3 +1007,413 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+    coordinates : :doc:`pf.Coordinates <pyfar:classes/pyfar.coordinates>` or
+        `sp.SamplingSphere <spharpy:classes/spharpy.samplings.coordinates>`
+        objects with sampling points for which the basis matrix is
+        calculated
+    basis_type : str, optional
+        Type of spherical harmonic basis, either ``'complex'`` or
+        ``'real'``. The default is ``'real'``.
+    normalization : str, optional
+        Normalization convention, either ``'n3d'``, ``'maxN'`` or
+         ``'sn3d'``. The default is ``'n3d'``.
```
ahms5, line 1104:
```suggestion
        ``'sn3d'``. The default is ``'n3d'``.
```
this space will fix the rendering in sphinx

spharpy/spherical.py
```diff
@@ -1005,3 +1007,413 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+                             "and 'fuma' are supported"
+                             )
+        if value == "fuma" and self.n_max > 3:
+            raise ValueError("n_max > 3 is not allowed with 'fuma' channel convention")
+        if value != self._channel_convention:
+            self._recompute_basis = True
+            self._recompute_basis_gradient = True
+            self._recompute_inverse = True
+            self._recompute_inverse_gradient = True
+        self._channel_convention = value
```
f-brinkmann, line 1296: Move inside if case

spharpy/spherical.py
```diff
@@ -1005,3 +1007,413 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+    @channel_convention.setter
+    def channel_convention(self, value):
+        """Set the channel order convention."""
+        if value not in ["acn", "fuma"]:
+            raise ValueError("Invalid channel convention, "
+                             "currently only 'acn' "
+                             "and 'fuma' are supported"
+                             )
+        if value == "fuma" and self.n_max > 3:
+            raise ValueError("n_max > 3 is not allowed with 'fuma' channel convention")
```
f-brinkmann, line 1290: Line too ling

spharpy/spherical.py
```diff
@@ -1005,3 +1007,413 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+    def normalization(self, value):
+        """Set the normalization convention."""
+        if value not in ["n3d", "maxN", "sn3d"]:
+            raise ValueError(
+                "Invalid normalization, "
+                "currently only 'n3d', 'maxN', 'sn3d' are "
+                "supported"
+            )
+        if value == "maxN" and self.n_max > 3:
+            raise ValueError("n_max > 3 is not allowed with 'maxN' normalization")
```
f-brinkmann, line 1313: Line too long

spharpy/spherical.py
```diff
@@ -1005,3 +1007,413 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        if value != self._normalization:
+            self._recompute_basis = True
+            self._recompute_basis_gradient = True
+            self._recompute_inverse = True
+            self._recompute_inverse_gradient = True
+        self._normalization = value
```
f-brinkmann, line 1319: move inside if case

ahms5 (commented):
looks fine, thank you, execpt the things which went wrong in the documentation. The suggested changes should fix all problems (hopefully) otherwise approved!

spharpy/spherical.py
```diff
@@ -1005,3 +1007,413 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        self._recompute_basis_gradient = True
+        self._recompute_inverse = True
+        self._recompute_inverse_gradient = True
+
+        self._basis = None
+        (self._basis_gradient_theta,
+         self._basis_gradient_phi) = (None, None)
+        self._basis_inv = None
+        (self._basis_inv_gradient_theta,
+         self._basis_inv_gradient_phi) = (None, None)
```
f-brinkmann, line 1156: I think we might not need this: If you set `self._recompute = True` already in the setter. All getters for this, e.g., `self.basis` could directly and only check `self._recompute` and don't need to check `if self._basis is None`

spharpy/spherical.py
```diff
@@ -1005,3 +1007,413 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+            self._recompute_basis = True
+            self._recompute_basis_gradient = True
+            self._recompute_inverse = True
+            self._recompute_inverse_gradient = True
+        self._normalization = value
+
+    @property
+    def basis(self):
+        """ndarray: Get the spherical harmonic basis matrix."""
+        if self._basis is None or self._recompute_basis:
```
f-brinkmann, line 1324: See comment in `__init__` check for `is None` maybe not needed.

spharpy/spherical.py
```diff
@@ -1005,3 +1007,413 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+            self._compute_basis()
+            self._recompute_basis = False
+        return self._basis
+
+    @property
+    def basis_gradient_theta(self):
+        """ndarray: Get the gradient of the basis matrix with
+        respect to theta."""
+        if (self._basis_gradient_theta is None or
```
f-brinkmann, line 1333: See comment in `__init__` check for `is None` maybe not needed.

spharpy/spherical.py
```diff
@@ -1005,3 +1007,413 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+                self._recompute_basis_gradient):
+            self._compute_basis_gradient()
+            self._recompute_basis_gradient = False
+        return self._basis_gradient_theta
+
+    @property
+    def basis_gradient_phi(self):
+        """ndarray: Get the gradient of the basis
+        matrix with respect to phi."""
+        if (self._basis_gradient_phi is None or
```
f-brinkmann, line 1343: See comment in `__init__` check for `is None` maybe not needed.

spharpy/spherical.py
```diff
@@ -1005,3 +1007,413 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        if self._basis is None:
+            self._compute_basis()
+        if self.inverse_transform is not None:
+            if self.inverse_transform not in ["pseudo_inverse",
+                                              "quadrature"]:
+                raise ValueError(
+                    "Invalid inverse transform type, "
+                    "currently only 'pseudo_inverse' "
+                    "and 'quadrature' are supported"
+                )
```
f-brinkmann, line 1411: This check should go to the setter for `inverse_transform` and should be removed here.

spharpy/spherical.py
```diff
@@ -1005,3 +1007,413 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+            self._recompute_inverse = False
+        return self._basis_inv
+
+    def _compute_inverse(self):
+        """
+        Compute the inverse basis matrix for the SphericalHarmonics class
+        """
+        if self._basis is None:
+            self._compute_basis()
+        if self.inverse_transform is not None:
```
f-brinkmann, line 1404: See comment in `__init__` maybe we do not initialize this with `None` and than also dont need to check for it here.

spharpy/spherical.py
```diff
@@ -1005,3 +1007,413 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+            if self.inverse_transform not in ["pseudo_inverse",
+                                              "quadrature"]:
+                raise ValueError(
+                    "Invalid inverse transform type, "
+                    "currently only 'pseudo_inverse' "
+                    "and 'quadrature' are supported"
+                )
+        elif self.inverse_transform is None:
+            raise ValueError("Inverse transform type not specified")
```
f-brinkmann, line 1413: See above. Maybe get rid of this.

spharpy/spherical.py
```diff
@@ -1005,3 +1007,413 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        if self.inverse_transform == "pseudo_inverse":
+            self._basis_inv = np.linalg.pinv(self._basis)
+        elif self.inverse_transform == "quadrature":
+            if self.weights is None:
+                self.coordinates.weights = calculate_sampling_weights(self.coordinates)
```
f-brinkmann, line 1418: See comment in `weights` getter. I would suggest to not silently compute sampling weights for the user. But maybe thats something to discuss during the weekly.

f-brinkmann (changes requested):
Thanks for continuing with this. Mostly small comments, but general things about initializing values with `None`, which might not be required and having multiple variables to check for recompute instead of only `self._recompute` (see detailed comments). We can maybe discuss the more general stuff during the next weekly before you implement this...

spharpy/spherical.py
xefonon: yes, because this description is already covered by the initialisation (parameter) doc. So it's like a duplicate here.

spharpy/spherical.py
f-brinkmann, line 1156: Do we need both? Instead of setting `recompute=True` we could set everything to `None` when changing parameters that require recompute.

spharpy/spherical.py
f-brinkmann: If we keep, it should be checked, if value.shape matches the shape of the coordinates

spharpy/spherical.py
```diff
@@ -1005,3 +1007,413 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
 
     return identity_matrix
+
+
+
+class SphericalHarmonics:
+    r"""
+    Compute spherical harmonic basis matrices, their inverses, and gradients.
+    The spherical harmonic Ynm is given by:
+    .. math::
```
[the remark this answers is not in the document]
xefonon: Resolving for a later PR

spharpy/spherical.py
xefonon: As we discussed, we shouldn't calculate the weights silently (or perhaps sneakily), but we should provide a way for the user to add their own weights in the class. It is typical in SH to have analytically estimated weights for various sampling schemes which currently we don't provide in spharpy. It is more intuitive imho to be able to set, or provide these weights directly to a SH class, and not implicitly via a coordinates object.

xefonon: Also if things like channel convention are re-set e.g. `sph_obj.channel_convention = 'fuma'`, `weights` would have to be re-set or re-computed anyway, this is not a part of `pf.coordinates` object, and will probably never make sense to be either. Right now with the change (and no sneaky computation of `weights` ), I set `self.weights` to `None` and the user will get a `ValueError` when they try to compute inverse.

spharpy/spherical.py
```diff
@@ -1005,3 +1007,413 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        n_max,
+        coordinates,
+        basis_type="real",
+        normalization="n3d",
+        channel_convention="acn",
+        inverse_transform=None,
+        phase_convention=None
+    ):
+        # hidden attributes
+        self._n_max = None
```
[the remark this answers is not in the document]
xefonon: Yes the hidden attribute initialisation is required for the setters. If `self._n_max` is not set, one get's an `AttributeError: 'SphericalHarmonics' object has no attribute '_n_max'` and so forth.

xefonon:
`Helloooooooo` and thanks for the reviews!! :)

Fresh updated code, @f-brinkmann @ahms5

spharpy/spherical.py
```diff
@@ -1005,3 +1004,402 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+            raise ValueError("Coordinates cannot be empty")
+        if value != self._coordinates:
+            self._reset_compute_attributes()
+            self._coordinates = value
+            # Reset weights if coordinates change
+            self._weights = value.weights
+
+    @property
+    def basis_type(self):
+        """Get the type of spherical harmonic basis."""
```
f-brinkmann, line 1233: See comment on getter/setter above

spharpy/spherical.py
```diff
@@ -1005,3 +1004,402 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        if value not in ["complex", "real"]:
+            raise ValueError("Invalid basis type, only "
+                             "'complex' and 'real' are supported")
+        if value != self._basis_type:
+            self._reset_compute_attributes()
+            self._basis_type = value
+
+    @property
+    def inverse_transform(self):
+        """Get the type of inverse transform."""
```
f-brinkmann, line 1250: See comment on getter/setter above

spharpy/spherical.py
```diff
@@ -1005,3 +1004,402 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+            "Allowed values are 'pseudo_inverse', 'quadrature', or None.")
+        if value == "quadrature" and self._weights is None:
+            raise ValueError("Weights must be set for quadrature inverse transform")        
+        if value != self._inverse_transform:
+            self._reset_compute_attributes()
+            self._inverse_transform = value
+
+    @property
+    def channel_convention(self):
+        """Get the channel ordering convention."""
```
f-brinkmann, line 1267: See comment on getter/setter above

spharpy/spherical.py
```diff
@@ -1005,3 +1004,402 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        if value == "fuma" and self.n_max > 3:
+            raise ValueError("n_max > 3 is not allowed with 'fuma' " \
+                            "channel convention")
+        if value != self._channel_convention:
+            self._reset_compute_attributes()
+            self._channel_convention = value
+
+    @property
+    def normalization(self):
+        """Get the normalization convention."""
```
f-brinkmann, line 1287: See comment on getter/setter above

spharpy/spherical.py
```diff
@@ -1005,3 +1004,402 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        if self._basis is None:
+            self._compute_basis()
+        if self.inverse_transform == "pseudo_inverse":
+            self._basis_inv = np.linalg.pinv(self._basis)
+        elif self.inverse_transform == "quadrature":
+            if self.weights is None:
+                raise ValueError("Weights must be set for quadrature inverse transform")
+            self._basis_inv = np.einsum('ij,i->ji', np.conj(self._basis), self.weights)
+        elif self.inverse_transform is None:
+            self._basis_inv = np.eye(self._basis.shape[1])
```
f-brinkmann, line 1392: Can this be None? The default is 'pseudo_inverse'. This case can probably be removed.

tests/test_sh_class.py
```diff
@@ -0,0 +1,128 @@
...
+from spharpy.samplings import gaussian, calculate_sampling_weights
+
+def test_sphharm_init():
+    coordinates = pf.Coordinates(1, 0, 0)
+    sph_harm = SphericalHarmonics(n_max=2, coordinates=coordinates)
+    assert sph_harm.n_max == 2
+    assert np.all(sph_harm.coordinates == coordinates)
+
+def test_sphharm_init_invalid_coordinates():
+    with pytest.raises(TypeError):
```
f-brinkmann, line 17: We usually match the error message to make sure that the correct error is raised. Can you update that for all tests?
```suggestion
    with pytest.raises(TypeError, match='some string'):
```

tests/test_sh_class.py
```diff
@@ -0,0 +1,128 @@
...
+import numpy as np
+import pyfar as pf
+from spharpy.spherical import SphericalHarmonics
+from spharpy.samplings import gaussian, calculate_sampling_weights
+
+def test_sphharm_init():
+    coordinates = pf.Coordinates(1, 0, 0)
+    sph_harm = SphericalHarmonics(n_max=2, coordinates=coordinates)
+    assert sph_harm.n_max == 2
+    assert np.all(sph_harm.coordinates == coordinates)
```
f-brinkmann, line 14: I would suggest to test the values of the optional parameters at least once as well, e.g., `sph_harm.basis_type == 'real'`

tests/test_sh_class.py
```diff
@@ -0,0 +1,128 @@
...
+def test_sphharm_compute_inverse_quad():
+    coordinates = gaussian(n_points=4)
+    weights = calculate_sampling_weights(coordinates)
+    coordinates.weights = weights
+    sh = SphericalHarmonics(2, coordinates, inverse_transform = 'quadrature')
+    assert sh.basis_inv is not None
+
+def test_sphharm_compute_inverse_pseudo_inv():
+    coordinates = gaussian(n_points= 5)
+    sh = SphericalHarmonics(2, coordinates, inverse_transform = 'pseudo_inverse')
```
f-brinkmann, line 45: line too long

tests/test_sh_class.py
```diff
@@ -0,0 +1,128 @@
...
+    assert sh.basis_inv is not None
+
+def test_compute_basis_caching():
+    n_max = 2
+    points = np.random.rand(10, 3)
+    coordinates = pf.Coordinates(points[:, 0], points[:, 1], points[:, 2], 'cart')
```
f-brinkmann, line 51: line too long

tests/test_sh_class.py
```diff
@@ -0,0 +1,128 @@
...
+        # set sph_harm to use 'fuma' channel convention
+        sph_harm.n_max = 2  # Invalid with default 'acn' and 'maxN'
+        sph_harm.channel_convention = "fuma"
+        sph_harm.n_max = 4  # Invalid with default 'acn' and 'maxN'
+    with pytest.raises(ValueError):
+        sph_harm.channel_convention = "acn"
+        sph_harm.n_max = 4
+        # set maxN normalization
+        sph_harm.normalization = "maxN"  # Invalid with n_max > 3
+def test_setter_phase_convention():
```
f-brinkmann, line 84: missing blank lines before def

f-brinkmann (changes requested):
Thanks for the update. I mostly had comments on the docstrings left

tests/test_sh_class.py
xefonon: Can you share your ruff settings, because my linter is not complaining?

spharpy/spherical.py
xefonon: Of course it's not necessary but there might be some use cases where you just want the inverse to do nothing, within the context of a placeholder.

tests/test_sh_class.py
f-brinkmann: I just checked. Line length of 79 is already checked on master, where we switched to pyproject.toml but not on dev. You can add a vertical line in VSCode to see lines that are too long: https://www.reddit.com/r/vscode/comments/evdtir/how_to_add_a_vertical_line_showing_that_ive/

spharpy/spherical.py
f-brinkmann: But can `self.inverse_transform` ever be `None`? I haven't seen how this could happen.

spharpy/spherical.py
```diff
@@ -1005,3 +1004,402 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        """
+        if value not in ["complex", "real"]:
+            raise ValueError("Invalid basis type, only "
+                             "'complex' and 'real' are supported")
+        if value != self._basis_type:
+            self._reset_compute_attributes()
+            self._basis_type = value
+
+    @property
+    def inverse_transform(self):
```
mberz, line 1249: I have two comments on this:

1. I just started to realize that this property actually is a property of the spherical sampling, rather than the spherical harmonics. It may be worthwhile to move the property to SamplingSphere. This would have some implications:
    i. Computing the inverse matrix for `pyfar.Coordinates` would not be possible, which may actually be beneficial, as they might not even contain points distributed on a sphere. Additionally, checks for empty weights etc. would be moved to SamplingSphere, simplifying the SphericalHarmonic class.
    ii. Computing `basis` remains unchanged.
    iii. The duplicate weights property issue would be solved.
2. To me the naming of the property is a bit misleading, as it does not give the inverse transform, but rather defines the operation to be performed when computing the inverse.

mberz (commented):
@xefonon Sorry, I can only comment and not do a full review as the PR was openend under my username.

I have one design comment and otherwise primarily have dostring comments Could you please improve spelling in error Messages when refering to specific class implementations, for example pyfarf.Coodinates instead of pyfar.coordinates.

The design comment is on the `inverse_transform` property, please see the respective comment. I'm not sure how much work it would be. Maybe it's even worth creating a separate PR for it. Could you maybe comment on that?

spharpy/spherical.py
```diff
@@ -1005,3 +1004,402 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        """array-like: Sampling weights for the quadrature transform.
+        Obtained from the coordinates object and are
+        normalized to a sum of 4*pi."""
+        if self._weights is not None:
+            return self._weights
+        elif hasattr(self.coordinates, 'weights') and self.coordinates.weights is not None:
+            return self.coordinates.weights
+        else:
+            raise ValueError("weights not set, please set them manually or use a "
+                             "pyfar.coordinates object with weights")
```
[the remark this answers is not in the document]
mberz: There are more occurences of issues like that. Please check :)

spharpy/spherical.py
```diff
@@ -782,13 +784,11 @@ def aperture_vibrating_spherical_cap(
         Radius of the sphere
     r_cap : double
         Radius of the vibrating cap
-
     Returns
     -------
-    A : ndarray, float
+    aperture : ndarray, float
```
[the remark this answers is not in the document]
xefonon: was a suggested change

spharpy/spherical.py
```diff
@@ -823,7 +823,7 @@ def aperture_vibrating_spherical_cap(
 
 def radiation_from_sphere(
         n_max,
-        rad_sphere,
+        r_sphere,
```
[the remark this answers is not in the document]
xefonon: `r_sphere` is more consistent with the rest of the `spherical.py` code than `rad_sphere`

spharpy/spherical.py
```diff
@@ -854,11 +854,10 @@ def radiation_from_sphere(
...
     k : ndarray, float
         Wave number
-    distance : float
-        Radial distance from the center of the sphere
     density_medium : float
         Density of the medium surrounding the sphere. Default is 1.2 for air.
-    speed_of_sound : float
+    distance : float
+        Radial distance from the center of the sphere
         Speed of sound in m/s
```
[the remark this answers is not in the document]
xefonon: This is part of an issue I mentioned after first commit, when I pulled `dev` changes and then  applied suggested changes afterwards. Thanks for spotting.

spharpy/spherical.py
```diff
@@ -877,7 +876,7 @@ def radiation_from_sphere(
     for n in range(n_max+1):
         hankel = _special.spherical_hankel(n, k*distance, kind=2)
         hankel_prime = _special.spherical_hankel(
-            n, k*rad_sphere, kind=2, derivative=True)
+            n, k*r_sphere, kind=2, derivative=True)
```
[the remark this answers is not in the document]
xefonon: see comment above

spharpy/spherical.py
```diff
@@ -1005,3 +1004,402 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        self.condon_shortley = condon_shortley
+
+    # Properties
+    @property
+    def condon_shortley(self):
+        """Set the Condon-Shortley phase term."""
+        return self._condon_shortley
+
+    @condon_shortley.setter
+    def condon_shortley(self, value):
```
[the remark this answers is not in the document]
xefonon: This is checked in `spherical_harmonic_basis`, `spherical_harmonic_basis_gradient_real`, etc. functions. It would be redundant.

spharpy/spherical.py
```diff
@@ -1005,3 +1004,402 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+
+    @property
+    def weights(self):
+        """array-like: Sampling weights for the quadrature transform.
+        Obtained from the coordinates object and are
+        normalized to a sum of 4*pi."""
+        if self._weights is not None:
+            return self._weights
+        elif hasattr(self.coordinates, 'weights') and self.coordinates.weights is not None:
+            return self.coordinates.weights
```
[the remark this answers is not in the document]
xefonon: Not necessarily, `self._weights` is None from a generic `pf.Coordinates` object. The second part of the `if` statement, e.g., `self.coordinates.weights is not None` is a requirement for weights to be set in the default state. And `self.weights` are only used in the `self.inverse_transform='quadrature'` case. In other words: they can't be `None` when the user requires a integration by quadrature.

spharpy/spherical.py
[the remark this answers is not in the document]
xefonon: This is the only one I could find like this.

spharpy/spherical.py
```diff
@@ -1005,3 +1004,402 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        if not isinstance(value, np.ndarray):
+            raise TypeError("weights must be a numpy array")
+        if value.shape != (self.coordinates.csize,):
+            raise ValueError(
+                f"weights shape {value.shape} does not match the shape of coordinates "
+                f"({self.coordinates.csize},)"
+            )
+        if (value != self._weights).any():
+            self._reset_compute_attributes()
+            self._weights = value
```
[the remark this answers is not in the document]
xefonon: but if that's the consensus I'm happy to change.

spharpy/spherical.py
xefonon: I'm not sure I understand the connection between spherical sampling and the inverse transformation of a spherical harmonics basis matrix, but maybe we need to discuss this a bit more. Regarding the naming property I'm fine with changing, do you have any suggestions?

spharpy/spherical.py
```diff
@@ -1005,3 +1004,402 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+            self._basis_inv = np.einsum('ij,i->ji', np.conj(self._basis), self.weights)
+        elif self.inverse_transform is None:
+            self._basis_inv = np.eye(self._basis.shape[1])
+        else:
+            raise ValueError(
+                "Invalid inverse transform type. "
+                "Allowed values are 'pseudo_inverse', 'quadrature', or None."
+            )
+
+    def _reset_compute_attributes(self):
```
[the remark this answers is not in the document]
xefonon: let me know what you think.

spharpy/spherical.py
```diff
@@ -1005,3 +1004,402 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        self._basis = None
+        self._basis_gradient_theta = None
+        self._basis_gradient_phi = None
+        self._basis_inv = None
+        self._basis_inv_gradient_theta = None
+        self._basis_inv_gradient_phi = None
```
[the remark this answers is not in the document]
xefonon: Good catch!

spharpy/spherical.py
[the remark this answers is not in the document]
mberz: See comment above.

spharpy/spherical.py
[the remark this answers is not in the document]
mberz: See comment above. 🤣

spharpy/spherical.py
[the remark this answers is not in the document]
mberz: Probably I was halucinating.

spharpy/spherical.py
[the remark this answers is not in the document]
f-brinkmann: +1 for this. We can document that in the docstring for the inverse method

spharpy/spherical.py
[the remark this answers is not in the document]
f-brinkmann: I would check it here nevertheless to provide direct feedback to the user when entering stuff. Otherwise someone might for example save a class object with an invalid value.

spharpy/spherical.py
xefonon: Sure I understand that the integration is inherent to the sampling, but when you approximate the integration with a least squares optimal solution (and therefore use a pseudo-inverse) there is no necessity in knowing the exact sampling of the spherical function.

xefonon: And since we are talking about discrete approximations via matrices of spherical functions then referring to `inverse` functions or the sort in a `SamplingSphere` or `SamplingBall` object is totally counter intuitive rather than having the inverse functionality in the actual spherical function object.

spharpy/spherical.py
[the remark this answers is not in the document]
xefonon: OK add checks here aswell

xefonon:
@mberz @f-brinkmann @ahms5  Ping!

spharpy/samplings/coordinates.py
```diff
@@ -310,3 +310,17 @@ def n_max(self, value):
             self._n_max = None
         else:
             self._n_max = int(value)
+
+    @property
```
f-brinkmann, line 314: Note: This is reviewed in #120 but also required here. This PR relies on the property but does not care how it is implemented.

spharpy/spherical.py
```diff
@@ -960,7 +963,7 @@ def radiation_from_sphere(
     ----------
     n_max : integer
         Maximal spherical harmonic order
-    r_sphere : float
+    rad_sphere : float
```
f-brinkmann, line 966: Should this be reverted as well?

spharpy/spherical.py
```diff
@@ -1115,3 +1118,385 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        self._condon_shortley = value
+
+    @property
+    def n_max(self):
+        """Get or set the spherical harmonic order."""
+        return self._n_max
+
+    @n_max.setter
+    def n_max(self, value):
+        """Set the spherical harmonic order."""
```
f-brinkmann, line 1287: I think you forgot to copy the docstring of the setter to the getter in this case

spharpy/spherical.py
```diff
@@ -1115,3 +1118,385 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        if self.normalization == "maxN" and value > 3:
+            raise ValueError("n_max > 3 is not allowed with 'maxN' "
+                             "normalization")
+        if int(value) != self._n_max:
+            self._reset_compute_attributes()
+            self._n_max = int(value)  # Cast to int for safety
+
+    @property
+    def coordinates(self):
+        """Get the coordinates object."""
```
f-brinkmann, line 1304: Getter and setter docstring should be identical and start with 'Get or set...'

spharpy/spherical.py
```diff
@@ -1115,3 +1118,385 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        return self._coordinates
+
+    @coordinates.setter
+    def coordinates(self, value):
+        """Set the coordinates object."""
```
f-brinkmann, line 1309: Getter and setter docstring should be identical and start with 'Get or set...'

spharpy/spherical.py
```diff
@@ -1115,3 +1118,385 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+
+    @property
+    def basis_type(self):
+        """Get or set the type of spherical harmonic basis."""
+        return self._basis_type
+
+    @basis_type.setter
+    def basis_type(self, value):
+        """
+        Set the type (real or complex) of spherical harmonic basis.
```
f-brinkmann, line 1329: Getter and setter docstring should be identical and start with 'Get or set...'

spharpy/spherical.py
```diff
@@ -1115,3 +1118,385 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+            self._basis_type = value
+
+    @property
+    def inverse_method(self):
+        """Get or set the type of inverse transform."""
+        return self._inverse_method
+
+    @inverse_method.setter
+    def inverse_method(self, value):
+        """Set the inverse transform type."""
```
f-brinkmann, line 1345: Getter and setter docstring should be identical and start with 'Get or set...'

spharpy/spherical.py
```diff
@@ -1115,3 +1118,385 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        if value not in allowed:
+            raise ValueError("Invalid inverse method. Allowed values are 'inverse', "
+                             "'quadrature', or 'auto'.")
+        # If using simple pf.Coordinates and not SamplingSphere, 'auto' is not allowed
+        if not isinstance(self.coordinates, SamplingSphere) and value == "auto":
+            raise ValueError("inverse_method='auto' is not allowed without SamplingSphere coordinates. " \
+            "Please specify 'pseudo_inverse' or 'quadrature'.")
+        # If using quadrature, weights must be set
+        if value == "quadrature" and self.coordinates.weights is None:
+            raise ValueError("Weights must be set for quadrature inverse method")
```
f-brinkmann, line 1356: You can use the new property now:
```suggestion
        if not hasattr(self, 'quadrature') or not self.quadrature:
            raise ValueError("'quadrature' requires coordinates in a SamplingSphere object and SampülingSphere.quadrature = True")
```

spharpy/spherical.py
```diff
@@ -1115,3 +1118,385 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        allowed = ["pseudo_inverse", "quadrature", "auto"]
+        if value not in allowed:
+            raise ValueError("Invalid inverse method. Allowed values are 'inverse', "
+                             "'quadrature', or 'auto'.")
+        # If using simple pf.Coordinates and not SamplingSphere, 'auto' is not allowed
+        if not isinstance(self.coordinates, SamplingSphere) and value == "auto":
+            raise ValueError("inverse_method='auto' is not allowed without SamplingSphere coordinates. " \
+            "Please specify 'pseudo_inverse' or 'quadrature'.")
+        # If using quadrature, weights must be set
+        if value == "quadrature" and self.coordinates.weights is None:
```
f-brinkmann, line 1355: This is not yet complicated and looking at what we discussed again, I think its quite convoluted. Maybe the late afternoon decision was not the best. Maybe this would be more straight forward:

- do not alow 'auto'
- quadrature requires SamplingSphere.quadrature = True

Any opinons?

spharpy/spherical.py
```diff
@@ -1115,3 +1118,385 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+            self._inverse_method = value
+
+    @property
+    def channel_convention(self):
+        """Get or set the channel ordering convention."""
+        return self._channel_convention
+
+    @channel_convention.setter
+    def channel_convention(self, value):
+        """Set the channel order convention."""
```
f-brinkmann, line 1368: Getter and setter docstring should be identical and start with 'Get or set...'

spharpy/spherical.py
```diff
@@ -1115,3 +1118,385 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+            self._channel_convention = value
+
+    @property
+    def normalization(self):
+        """Get or set the normalization convention."""
+        return self._normalization
+
+    @normalization.setter
+    def normalization(self, value):
+        """Set the normalization convention."""
```
f-brinkmann, line 1388: Getter and setter docstring should be identical and start with 'Get or set...'

spharpy/spherical.py
```diff
@@ -1115,3 +1118,385 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        if value == "maxN" and self.n_max > 3:
+            raise ValueError("n_max > 3 is not allowed with " \
+                            "'maxN' normalization")
+        if value != self._normalization:
+            self._reset_compute_attributes()
+            self._normalization = value
+
+    @property
+    def basis(self):
+        """Get or set the spherical harmonic basis matrix."""
```
f-brinkmann, line 1404: This does not have a setter right? In this case I would suggest
```suggestion
        """Get the spherical harmonic basis matrix."""
```

spharpy/spherical.py
```diff
@@ -1115,3 +1118,385 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        if self._basis is None:
+            self._compute_basis()
+        return self._basis
+
+    @property
+    def basis_gradient_theta(self):
+        """Get the gradient of the basis matrix with
+        respect to theta."""
```
f-brinkmann, line 1412: Just code styling - it fits a single line:

```suggestion
        """Get the gradient of the basis matrix with respect to theta."""
```

spharpy/spherical.py
```diff
@@ -1115,3 +1118,385 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        if self._basis_gradient_theta is None:
+            self._compute_basis_gradient()
+        return self._basis_gradient_theta
+
+    @property
+    def basis_gradient_phi(self):
+        """Get or set the gradient of the basis
+        matrix with respect to phi."""
```
f-brinkmann, line 1420: As above: Single line and getter only:
```suggestion
        """Get the gradient of the basis matrix with respect to phi."""
```

spharpy/spherical.py
```diff
@@ -1115,3 +1118,385 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        if self.inverse_method == "pseudo_inverse":
+            self._basis_inv = np.linalg.pinv(self._basis)
+        elif self.inverse_method == "quadrature":
+            self._basis_inv = np.einsum('ij,i->ji', np.conj(self._basis), 
+                                        self.coordinates.weights)
+        else:
+            raise ValueError(
+                "Invalid inverse transform method. "
+                "Allowed values are 'pseudo_inverse', 'quadrature', or 'auto'."
+            )
```
f-brinkmann, line 1493: I think this case should be deleted, since the value of `inverse_method` is checked in its setter.

f-brinkmann (changes requested):
Thanks for the update. Mostly smaller docstring comments. Unfortunately, I again opened a discussion on how to handle ìnverse_method` because I wondered if our decision from Friday was good or not. I'll wirte something about that in slack

ahms5 (dismissed):
I think after including 'auto' in the inverse calculation we are done.

spharpy/spherical.py
```diff
@@ -1115,3 +1118,385 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        Compute the inverse basis matrix for the SphericalHarmonics class
+        """
+        if self._basis is None:
+            self._compute_basis()
+        if not isinstance(self.coordinates, SamplingSphere):
+            warnings.warn("Using a simple pf.Coordinates object for inverse_basis." \
+            " Make sure this is intended.")
+        if self.inverse_method == "pseudo_inverse":
+            self._basis_inv = np.linalg.pinv(self._basis)
+        elif self.inverse_method == "quadrature":
```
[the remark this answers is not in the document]
xefonon: Sorry I missed the weekly but I thought the idea was to remove 'auto', let me know if I should revert

xefonon:
Hey @mberz @f-brinkmann @ahms5, I've updated the `SphericalHarmonics` class.

spharpy/spherical.py
```diff
@@ -1104,3 +1107,373 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+            self.coordinates.quadrature = True
```
ahms5, line 1469:
```suggestion
```

spharpy/spherical.py
```diff
@@ -1104,3 +1107,373 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+    def _compute_inverse(self):
+        """
+        Compute the inverse basis matrix for the SphericalHarmonics class
+        """
+        if self._basis is None:
+            self._compute_basis()
+        if not isinstance(self.coordinates, SamplingSphere):
+            warnings.warn("Using a simple pf.Coordinates object for inverse_basis." \
+            " Make sure this is intended.")
+        if self.inverse_method == "pseudo_inverse":
```
ahms5, line 1466:
```suggestion
        inverse_method = self.inverse_method
        if self.inverse_method == 'auto':
            if not isinstance(self.coordinates, SamplingSphere):
                raise ValueError("'inverse_method' cannot be 'auto', if 'coordinates' is not 'SamplingSphere'.")
                inverse_method = 'quadrature' if self.coordinates.quadrature else 'pseudo_inverse'
        if inverse_method == "pseudo_inverse":
```

spharpy/spherical.py
```diff
@@ -1104,3 +1107,373 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+            self._basis_inv = np.linalg.pinv(self._basis)
+        elif self.inverse_method == "quadrature":
```
ahms5, line 1468:
```suggestion
        elif inverse_method == "quadrature":
```

spharpy/spherical.py
```diff
@@ -1104,3 +1107,373 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        - ``'acn'``: Follows the Ambisonic Channel Number (ACN) convention.
+        - ``'fuma'``: Follows the Furse-Malham (FuMa) convention.
+        (FuMa is only supported up to 3rd order)
+
+    - inverse_method: Defines the type of inverse transform.
+
+        - ``'pseudo_inverse'``: Uses the Moore-Penrose pseudo-inverse
+         for the inverse transform.
+        - ``'quadrature'``: Uses quadrature for the inverse transform.
+        - ``None``: No inverse transform is applied and basis is returned.
```
ahms5, line 1193:
```suggestion
        - ``'auto'``: ``'quadrature'`` if `coordinates.quadrature` 
           is True otherwise ``'quadrature'``. If `coordinates` is not 
           SamplingSphere an error is returned.


```

spharpy/spherical.py
```diff
@@ -1104,3 +1107,373 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        Normalization convention, either ``'n3d'``, ``'maxN'`` or
+        ``'sn3d'``. The default is ``'n3d'``.  
+        (maxN is only supported up to 3rd order)
+    channel_convention : str, optional
+        Channel ordering convention, either ``'acn'`` or ``'fuma'``.
+        The default is ``'acn'``.
+        (FuMa is only supported up to 3rd order)
+    inverse_method : str, optional
+        Inverse transform type, either ``'pseudo_inverse'`` or
+        ``'quadrature'``. The default is "pseudo_inverse".
```
ahms5, line 1215:
```suggestion
        Inverse transform type, either ``'auto'``, ``'pseudo_inverse'`` or
        ``'quadrature'``. The default is "auto".
```

spharpy/spherical.py
```diff
@@ -1104,3 +1107,373 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+    """
+
+    def __init__(
+        self,
+        n_max,
+        coordinates,
+        basis_type="real",
+        normalization="n3d",
+        channel_convention="acn",
+        inverse_method="pseudo_inverse",
```
ahms5, line 1231:
```suggestion
        inverse_method="auto",
```

tests/test_sh_class.py
```diff
@@ -0,0 +1,137 @@
...
+
+def test_sphharm_compute_inverse_pseudo_inv():
+    coordinates = gaussian(n_points= 5)
+    sh = SphericalHarmonics(2, coordinates, 
+                            inverse_method = 'pseudo_inverse')
+    assert sh.basis_inv is not None
+
+def test_compute_basis_caching():
+    n_max = 2
+    points = np.random.randint(4, 10)
```
ahms5, line 58:
```suggestion
    rng = np.random.default_rng()
    points = rng.integers(4, 10)
```

this is giving a ruff warning, since it is deprecated in numpy

ahms5 (commented):
Almost there, thank you!

[here](https://github.com/pyfar/spharpy/pull/101#issue-2866950885) we summarised the outcome of the last discussion, deciding on a compromise between the two ideas. The user is granted a useful default behaviour if they created the sampling with Spharpy (using the SamplingSphere.quatdrature information). If the user wants to override the default behaviour, they can use inverse_method.

f-brinkmann (commented):
@xefonon, @ahms5, @mberz @tluebeck , I'm sorry but I can't let go of the handling for `quadrature` and think we should reconsider our decision. Because we have a basis matrix Y and weights w, we have a clear mathematical criterion to check if we have a quadrature sampling, which is to test if `Y.T @ diag(w) @ Y` equals the identity matrix. This would mean that `SpharicalHarmonics.quadrature` could be a getter only property that checks

```python
import spharpy as sy
import numpy as np
import numpy.testing as npt

# generate a sampling
quadrature = True
if quadrature:
    sampling = sy.samplings.gaussian(n_max=3)
else:
    sampling = sy.samplings.hyperinterpolation(n_max=3)

# generate SH object
sh = sy.spherical.SphericalHarmonics(n_max=3, coordinates=sampling)

# test if basis is quadrature
n_coeffs = (sh.n_max + 1)**2
basis = sh.basis[:, :n_coeffs]
quadrature_test = basis.T @ np.diag(sh.coordinates.weights) @ basis
diag = np.diag(np.ones(n_coeffs))

# can not use relative tolerance when testing for identity matrix due to zeros
decimal = 10  # precision could be discussed or a class paramneter
npt.assert_almost_equal(quadrature_test, diag, decimal)
```

This would make things easier and more clear I think

- `quadrature` could be removed from init
- if setting `inverse_transform_method = 'quadrature'` the setter checks `SphericalHarmonics.quadrature` and raises an error if the criterion is not met.

tluebeck (changes requested):
In addition to the others' comments, I just found something in the condon_shortley setter.

f-brinkmann:
@xenofon, we discussed the following in the weekly

- the `inverse_method` property should be realized as mentioned in the pull request description at the top of the page but it should always return `quadrature` or `pseudo_inverse`. That if the user does `inverse_method='auto'`, the setter checks and writes the correct value.

spharpy/spherical.py
```diff
@@ -1104,3 +1107,397 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        if self._basis is None:
+            self._compute_basis()
+        _inv_flag = self.inverse_method
+        # warn when falling back to pseudo-inverse on a plain pf.Coordinates
+        if _inv_flag == "pseudo_inverse" and not isinstance(self.coordinates, SamplingSphere):
+            import warnings
+            warnings.warn(
+                "Coordinates is not a SamplingSphere, " \
+                "using Moore–Penrose pseudo-inverse instead of quadrature."
+            )
```
ahms5, line 1488:
```suggestion
```
not sure if this can happen, since we raise an error if we dont have a SamplingSphere and quadrature is set. So I think we can remove it here.

ahms5: If we want to keep it, we can import the warning at the top of the file instead

ahms5 (commented):
Nice, thank you. Just small thinks which I have noticed.

tluebeck (approved):
The code looks fine to me. The only issue is that there are multiple lines that are too long and other Flake8 errors, or am I wrong? If those are fixed, I approve.

spharpy/spherical.py
```diff
@@ -1104,3 +1107,397 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        return self._inverse_method
+
+    @inverse_method.setter
+    def inverse_method(self, value):
+        """Get or set the inverse transform type."""
+        # If the user passes "auto", require SamplingSphere and resolve it
+        if isinstance(value, str) and value == "auto":
+            if not isinstance(self.coordinates, SamplingSphere):
+                raise ValueError("'auto' is only valid if `coordinates` is a SamplingSphere.")
+            if self.coordinates.quadrature:
```
f-brinkmann, line 1349: I think this will raise an error for Coordinates objects because they dont have the property. This should fix it

```suggestion
            if type(coordinates) is SamplingSohere and self.coordinates.quadrature:
```

spharpy/spherical.py
```diff
@@ -1104,3 +1107,397 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+                value = "quadrature"
+            else:
+                value = "pseudo_inverse"
+        elif value == "quadrature":
+            if not isinstance(self.coordinates, SamplingSphere):
+                raise ValueError("'quadrature' requires `coordinates` to be a SamplingSphere.")
```
f-brinkmann, line 1355: Should this also check the quadrature attribute?

```suggestion
            if not isinstance(self.coordinates, SamplingSphere) or not coordinates.quadrature:
                raise ValueError("'quadrature' requires `coordinates` to be a SamplingSphere and coordinates.quadrature to be True.")
```

spharpy/spherical.py
```diff
@@ -1104,3 +1107,397 @@ def sph_identity_matrix(n_max, type='n-nm'):
...
+        Normalization convention, either ``'n3d'``, ``'maxN'`` or
+        ``'sn3d'``. The default is ``'n3d'``.  
+        (maxN is only supported up to 3rd order)
+    channel_convention : str, optional
+        Channel ordering convention, either ``'acn'`` or ``'fuma'``.
+        The default is ``'acn'``.
+        (FuMa is only supported up to 3rd order)
+    inverse_method : str, optional
+        Inverse transform type, either ``'auto'``, ``'pseudo_inverse'`` or  
+        ``'quadrature'``. The default is ``'auto'``.
```
f-brinkmann, line 1219: We should explain what 'auto' does. This assues, that we will move the mathematical descriptions and additional information to the beginning of the SH classes documentation.

```suggestion
        ``'quadrature'``. The default is ``'auto'``, which uses ``'quadrature'```
        if possible and ``'pseudo_inverse'`` otherwise (see above for details).
```

f-brinkmann (changes requested):
Thanks for the effort. I only had three minor comments left and agree with the suggestions from ahms5 and tluebeck

spharpy/spherical.py
xefonon: done

spharpy/spherical.py
xefonon: fixed

spharpy/spherical.py
xefonon: done

xefonon:
> The code looks fine to me. The only issue is that there are multiple lines that are too long and other Flake8 errors, or am I wrong? If those are fixed, I approve.

will fix with merge from develop branch

spharpy/spherical.py
```diff
@@ -1341,6 +1342,7 @@ def inverse_method(self):
 
     @inverse_method.setter
     def inverse_method(self, value):
+        from spharpy import SamplingSphere
```
f-brinkmann, line 1345: Can you please move this up to the imports at the top of the file?

f-brinkmann (approved):
Made one tiny comment on moving an import to the top of the file. Otherwise approved. Lets merge :)

spharpy/spherical.py
f-brinkmann: Ah, ok. For me it worked to `import spharpy as sy`. Maybe cleaner than an import somewhere in the file - not sure if ruff would complain about that.

xefonon: OK that worked! 👍

PR #102, opened 2025-02-20
Implement different normalization and phase conventions for spherical harmonic basis matrices

mberz:
Taken from #82


spharpy/spherical.py
```diff
@@ -352,40 +379,46 @@ def spherical_harmonic_basis_gradient(n_max, coords):
...
-    n_points = coords.csize
+
+    """  # noqa: 501
+    if not isinstance(coordinates, pf.Coordinates):
+        axis = np.where(coordinates.shape == 3)[0][0]
+        if axis == 0:
+            coordinates = coordinates.T
+        coordinates = pf.Coordinates(coordinates[:, 0],
+                                     coordinates[:, 1],
+                                     coordinates[:, 2])
```
mberz, line 395: I don't think we should do this transform. We should stick to allowing only pf.Coordinates or Samplings here.

spharpy/spherical.py
```diff
@@ -282,35 +302,41 @@ def spherical_harmonic_basis(n_max, coords):
...
+        )
+        if normalization == "sn3d":
+            basis[:, acn] *= n3d_to_sn3d_norm(order)
+        elif normalization == "maxN":
+            basis[:, acn] *= n3d_to_maxn(acn)
+        if phase_convention is None:
+            # Condon-Shortley phase term is already included in
+            # the special.spherical_harmonic function
+            # so need to divide by (-1)^m
+            basis[:, acn] /= (-1) ** float(degree)
```
mberz, line 332: Couldn't these re-normalization and re-sorting operations be moved out of the loop? This should be much faster.

spharpy/spherical.py
```diff
@@ -419,22 +456,32 @@ def spherical_harmonic_basis_real(n_max, coords):
...
```
mberz, line 479: Same comment as above, if possible, please move the re-normalization and sorting out of the loop.

mberz, line 479: I think the spherical_harmonic_real function does not include the CS phase. Please check the issue I created on this and compare.
Technically, there's no difference between multiplying with (-1)^m or dividing by, but the comment is not correct then ;)

spharpy/spherical.py
```diff
@@ -482,11 +530,18 @@ def spherical_harmonic_basis_gradient_real(n_max, coords):
...
-    theta = coords.colatitude
-    phi = coords.azimuth
+    """  # noqa: 501
+    if not isinstance(coordinates, pf.Coordinates):
+        axis = np.where(coordinates.shape == 3)[0][0]
+        if axis == 0:
+            coordinates = coordinates.T
+        coordinates = pf.Coordinates(coordinates[:, 0],
+                                     coordinates[:, 1],
+                                     coordinates[:, 2])
```
mberz, line 540: Same comment as above. I don't think we should do auto-transforming coordinates.,

tests/test_spherical_harmonics.py
```diff
@@ -40,24 +49,36 @@ def test_spherical_harmonic_n10(make_coordinates, implementation):
...
-    np.testing.assert_allclose(basis, reference, atol=1e-13)
+    Y = np.genfromtxt(f'./tests/data/Y_real_{phase_convention}_'
+                      f'{normalization}_{channel_convention}.csv',
+                      dtype=float,
+                      delimiter=',')
+    basis = sh.spherical_harmonic_basis_real(n_max, coords,
+                                             normalization,
+                                             channel_convention,
+                                             phase_convention=phase_convention)
+    np.testing.assert_allclose(basis, Y, atol=1e-13)
```
mberz, line 81: Could you also for the benefit of my sanity keep a test against the old files. Just so we're sure that we don't accidentally change the default normaltion and phase conventions of the files? ;)

spharpy/spherical.py
```diff
@@ -282,35 +302,41 @@ def spherical_harmonic_basis(n_max, coords):
...
+        if phase_convention is None:
+            # Condon-Shortley phase term is already included in
+            # the special.spherical_harmonic function
+            # so need to divide by (-1)^m
+            basis[:, acn] /= (-1) ** float(degree)
     return basis
 
 
-def spherical_harmonic_basis_gradient(n_max, coords):
+def spherical_harmonic_basis_gradient(n_max, coordinates):
```
mberz, line 336: You did not implement the phase and normalization conventions for the gradients. Is there a reason for this?
Did you want to keep this for a separate PR?

spharpy/spherical.py
```diff
@@ -271,9 +279,21 @@ def spherical_harmonic_basis(n_max, coords):
...
-    coordinates : :doc:`pf.Coordinates <pyfar:classes/pyfar.coordinates>`
-        Coordinate object with sampling points for which the basis matrix is
+    coordinates : :doc:`pf.Coordinates <pyfar:classes/pyfar.coordinates>` or
+        `sp.SamplingSphere <spharpy:classes/spharpy.samplings.coordinates>`
+        objects with sampling points for which the basis matrix is
         calculated
+    normalization : str, optional
+        Normalization convention, either 'n3d', 'maxN' or 'sn3d'.
+        The default is 'n3d'.
+        (maxN is only supported up to 3rd order)
```
mberz, line 289: Since this is only valid up to third order we should raise an error on invalid inputs.

spharpy/spherical.py
```diff
@@ -271,9 +279,21 @@ def spherical_harmonic_basis(n_max, coords):
...
+    channel_convention : str, optional
+        Channel ordering convention, either 'acn' or 'fuma'.
+        The default is 'acn'.
+        (FuMa is only supported up to 3rd order)
```
mberz, line 293: Since this is only valid up to third order we should raise an error on invalid inputs.

mberz (commented):
Thanks for cleaning this up and implementing the tests.
I can only comment since I opened the PR ;)

I have a couple of comments. Some I only noted in single places, but apply to all functions. It's probably easy to spot.

Could you especially please add a test against the old reference files to ensure that the new "default" settings still agree with the old behavior of the functions.

spharpy/spherical.py
tluebeck: It was a leftover from the old branch which I did not touch. But I removed it now.

spharpy/spherical.py
tluebeck: Same as above. It is removed now.

tests/test_spherical_harmonics.py
tluebeck: Sure, I added test against your old n10 files for both, real and complex. However, the new default for Condon-Shortley is None, so we explicitly need to test against phase_convention='Condon-Shortley', not the default.. Probably we should discuss at this point if we want to have "Condon-Shortley" as the default, to be consistent with old implementations?

spharpy/spherical.py
tluebeck: done

spharpy/spherical.py
tluebeck: done

spharpy/spherical.py
```diff
@@ -248,17 +249,24 @@ def n3d_to_sn3d_norm(n):
...
+        Y_n^m(\theta, \phi) =  CS_m N_{nm} P_{nm}(cos(\theta)) e^{im\phi}
 
-        Y_n^m(\theta, \phi) = \sqrt{\frac{2n+1}{4\pi}
-        \frac{(n-m)!}{(n+m)!}} P_n^m(\cos \theta) e^{i m \phi}
+    where:
+    - $n$ is the degree
+    - $m$ is the order
+    - $P_{nm}$ is the associated Legendre function
+    - $N_{nm}$ is the normalization term
+    - $CS_m$ is the Condon-Shortley phase term
```
f-brinkmann, line 267: Can we link somewhere for an analytical description of these? For the last, we could also write CS_m = (-1)^m directly here.

spharpy/spherical.py
f-brinkmann: +1 for this. I think _special.spherical_harmonic can also generate the entire matrix without being called in the loop. Could be a speed-up for high orders and large coordinates objects

spharpy/special.py
```diff
@@ -159,6 +159,8 @@ def _spherical_hankel_derivative(n, z, kind):
 
 def spherical_harmonic(n, m, theta, phi):
     """The spherical harmonics of order n and degree m.
+    The spherical harmonic functions are fully normalized (N3D) and
```
ahms5, line 162:
```suggestion

    The spherical harmonic functions are fully normalized (N3D) and
```
very picky suggestion^^

spharpy/special.py
```diff
@@ -195,7 +202,8 @@ def spherical_harmonic_real(n, m, theta, phi):
     r"""Real valued spherical harmonic function of order n and degree m
     evaluated at the angles theta and phi.
     The spherical harmonic functions are fully normalized (N3D) and follow
-    the AmbiX phase convention [#]_.
+    the AmbiX phase convention which does not include the Condon-Shortley
+    phase [#]_.
```
ahms5, line 206:
```suggestion
    the AmbiX phase convention, which does not include the Condon-Shortley
    phase [#]_.
```

spharpy/spherical.py
```diff
@@ -248,17 +249,24 @@ def n3d_to_sn3d_norm(n):
...
-def spherical_harmonic_basis(n_max, coords):
+def spherical_harmonic_basis(
+        n_max, coordinates, normalization="n3d", channel_convention="acn",
+        phase_convention='Condon-Shortley'):
     r"""
     Calculates the complex valued spherical harmonic basis matrix.
-
-    The spherical harmonic functions are fully normalized (N3D) and include the
-    Condon-Shotley phase term :math:`(-1)^m` [#]_, [#]_.
+    See also :func:`spherical_harmonic_basis_real`.
```
ahms5, line 257:
```suggestion

    See [#]_ and [#]_ for details.
    See also :py:func:`spherical_harmonic_basis_real`.
```
we need to refer the references, otherwise the doc build on ci will fail

f-brinkmann (changes requested):
Found only a few minor things. I generally wonder where we give a complete analytical description of the SHs? Is that given in the pull related to the SH-class?

spharpy/spherical.py
xefonon: With the same logic we could make this "jit"-able, but it's a bit of an overkill.

spharpy/spherical.py
```diff
@@ -248,17 +249,24 @@ def n3d_to_sn3d_norm(n):
...
 
     .. math::
+        Y_n^m(\theta, \phi) =  CS_m N_{nm} P_{nm}(cos(\theta)) e^{im\phi}
 
-        Y_n^m(\theta, \phi) = \sqrt{\frac{2n+1}{4\pi}
-        \frac{(n-m)!}{(n+m)!}} P_n^m(\cos \theta) e^{i m \phi}
+    where:
+    - $n$ is the degree
```
ahms5, line 263:
```suggestion

    - :math:`n` is the degree
```
like this, the empty line should also fix the bulletpoints

spharpy/spherical.py
```diff
@@ -271,9 +279,21 @@ def spherical_harmonic_basis(n_max, coords):
...
+        Normalization convention, either 'n3d', 'maxN' or 'sn3d'.
+        The default is 'n3d'.
+        (maxN is only supported up to 3rd order)
+    channel_convention : str, optional
+        Channel ordering convention, either 'acn' or 'fuma'.
+        The default is 'acn'.
+        (FuMa is only supported up to 3rd order)
+    phase_convention : string or None, optional
+        Whether to include the Condon-Shortley phase term.
+        The default is 'Condon-Shortley'.
```
ahms5, line 296: we should test if no other string than 'Condon-Shortley' is used

spharpy/spherical.py
```diff
@@ -282,35 +302,50 @@ def spherical_harmonic_basis(n_max, coords):
...
+            basis[:, acn] /= (-1) ** float(degree)
     return basis
 
 
-def spherical_harmonic_basis_gradient(n_max, coords):
+def spherical_harmonic_basis_gradient(n_max, coordinates, normalization="n3d",
+                                      channel_convention="acn",
+                                      phase_convention='Condon-Shortley'):
     r"""
     Calulcates the unit sphere gradients of the complex spherical harmonics.
```
ahms5, line 347:
```suggestion
    Calculates the unit sphere gradients of the complex spherical harmonics.
```

spharpy/spherical.py
```diff
@@ -339,9 +374,21 @@ def spherical_harmonic_basis_gradient(n_max, coords):
     ----------
     n_max : int
         Spherical harmonic order
-    coordinates : :doc:`pf.Coordinates <pyfar:classes/pyfar.coordinates>`
-        Coordinate object with sampling points for which the basis matrix is
+    coordinates : :doc:`pf.Coordinates <pyfar:classes/pyfar.coordinates>` or
+        `sp.SamplingSphere <spharpy:classes/spharpy.samplings.coordinates>`
```
ahms5, line 378: see above

spharpy/spherical.py
ahms5, line 347: there is a typo the first word

spharpy/spherical.py
```diff
@@ -339,9 +374,21 @@ def spherical_harmonic_basis_gradient(n_max, coords):
...
+        Normalization convention, either 'n3d', 'maxN' or 'sn3d'.
+        The default is 'n3d'.
+        (maxN is only supported up to 3rd order)
+    channel_convention : str, optional
+        Channel ordering convention, either 'acn' or 'fuma'.
+        The default is 'acn'.
+        (FuMa is only supported up to 3rd order)
+    phase_convention : string or None, optional
+        Whether to include the Condon-Shortley phase term.
+        The default is 'Condon-Shortley'.
```
ahms5, line 391: we should test if no other string than 'Condon-Shortley' is used

spharpy/spherical.py
```diff
@@ -248,17 +249,24 @@ def n3d_to_sn3d_norm(n):
     return 1 / np.sqrt(2 * n + 1)
 
 
-def spherical_harmonic_basis(n_max, coords):
+def spherical_harmonic_basis(
+        n_max, coordinates, normalization="n3d", channel_convention="acn",
+        phase_convention='Condon-Shortley'):
     r"""
     Calculates the complex valued spherical harmonic basis matrix.
```
ahms5, line 256: type calculates

spharpy/spherical.py
ahms5: same here, we need to ref the other reference

spharpy/spherical.py
```diff
@@ -352,40 +399,65 @@ def spherical_harmonic_basis_gradient(n_max, coords):
...
 
     return grad_theta, grad_phi
 
 
-def spherical_harmonic_basis_real(n_max, coords):
+def spherical_harmonic_basis_real(
+        n_max, coordinates, normalization="n3d", channel_convention="acn",
+        phase_convention=None):
     r"""
     Calculates the real valued spherical harmonic basis matrix.
```
ahms5, line 459: Calculates*

spharpy/spherical.py
```diff
@@ -352,40 +399,65 @@ def spherical_harmonic_basis_gradient(n_max, coords):
...
-
-    The spherical harmonic functions are fully normalized (N3D) and follow
-    the AmbiX phase convention [#]_.
+    See also :func:`spherical_harmonic_basis`.
```
ahms5, line 460:
```suggestion
    See also :py:func:`spherical_harmonic_basis`.
```

spharpy/spherical.py
xefonon: https://mathworld.wolfram.com/SphericalHarmonic.html

spharpy/spherical.py
```diff
@@ -396,21 +468,25 @@ def spherical_harmonic_basis_real(n_max, coords):
...
+        Normalization convention, either 'n3d', 'maxN' or 'sn3d'.
+        The default is 'n3d'.
+        (maxN is only supported up to 3rd order)
+    channel_convention : str, optional
+        Channel ordering convention, either 'acn' or 'fuma'.
+        The default is 'acn'.
+        (FuMa is only supported up to 3rd order)
+    phase_convention : string or None, optional
+        Whether to include the Condon-Shortley phase term.
+        The default is None.
```
ahms5, line 489: does it make sence to have different defaults for the "same" parameter? 
we should also highlight the possible opitons. 
if we wouldnt add more options in future we could also think about a boolean input ike ```apply_cs_phase```

spharpy/spherical.py
```diff
@@ -471,9 +567,21 @@ def spherical_harmonic_basis_gradient_real(n_max, coords):
     ----------
     n_max : int
         Spherical harmonic order
-    coordinates : :doc:`pf.Coordinates <pyfar:classes/pyfar.coordinates>`
-        Coordinate object with sampling points for which the basis matrix is
+    coordinates : :doc:`pf.Coordinates <pyfar:classes/pyfar.coordinates>` or
```
ahms5, line 570: see above

spharpy/spherical.py
```diff
@@ -419,22 +495,42 @@ def spherical_harmonic_basis_real(n_max, coords):
...
     return basis
 
 
-def spherical_harmonic_basis_gradient_real(n_max, coords):
+def spherical_harmonic_basis_gradient_real(n_max, coordinates,
+                                           normalization="n3d",
+                                           channel_convention="acn",
+                                           phase_convention=None):
     r"""
     Calulcates the unit sphere gradients of the real valued spherical hamonics.
```
ahms5, line 535: Calculates harmonics*

spharpy/spherical.py
```diff
@@ -471,9 +567,21 @@ def spherical_harmonic_basis_gradient_real(n_max, coords):
...
+        Normalization convention, either 'n3d', 'maxN' or 'sn3d'.
+        The default is 'n3d'.
+        (maxN is only supported up to 3rd order)
+    channel_convention : str, optional
+        Channel ordering convention, either 'acn' or 'fuma'.
+        The default is 'acn'.
+        (FuMa is only supported up to 3rd order)
+    phase_convention : string or None, optional
+        Whether to include the Condon-Shortley phase term.
+        The default is None.
```
ahms5, line 584: see above

ahms5 (commented):
Thank you for taking care! just checked the docs

spharpy/spherical.py
tluebeck: Are the two references not enough information for the analytical description or do we need the additional link?

f-brinkmann: Sorry that was my confusion - I forgot that we have a complete description as part of the SH-Class implementation

spharpy/spherical.py
f-brinkmann: Yes we need :math:`` here and everywhere below. You could search for '$' everywhere and replace it...

spharpy/spherical.py
```diff
@@ -271,9 +282,21 @@ def spherical_harmonic_basis(n_max, coords):
     ----------
     n_max : integer
         Spherical harmonic order
-    coordinates : :doc:`pf.Coordinates <pyfar:classes/pyfar.coordinates>`
-        Coordinate object with sampling points for which the basis matrix is
+    coordinates : :doc:`pf.Coordinates <pyfar:classes/pyfar.coordinates>` or `sp.SamplingSphere <spharpy:classes/spharpy.samplings.coordinates>`
```
f-brinkmann, line 285: Here and at some other places, lines are too long - can you check flake8 again, please?

spharpy/spherical.py
```diff
@@ -271,9 +282,21 @@ def spherical_harmonic_basis(n_max, coords):
...
+        The default is 'n3d'.
+        (maxN is only supported up to 3rd order)
+    channel_convention : str, optional
+        Channel ordering convention, either 'acn' or 'fuma'.
+        The default is 'acn'.
+        (FuMa is only supported up to 3rd order)
+    condon_shortley : bool or str, optional
+        Whether to include the Condon-Shortley phase term. If True or 'auto',
+        Condon-Shortley is included, if False it is not included.
+        The default is 'auto'.
```
f-brinkmann, line 299: I understood this in a different way: In the basic functions we have this as a bool with the reasonable default, which is `True` for the complex valued and `False` (?) for the real valued bases.

In the SH class in the other pull, we will have `'auto'` as a default, which will pick `True` or `False` depending on the basis type (complex or real).

Does that make sense to you?

f-brinkmann (changes requested):

spharpy/spherical.py
tluebeck: Yes, generally makes sense to me, I just understood it the other way round. I don't have a strong opinion which method should get the 'auto' option. I am just wondering if it wouldn't be nice to also have it in the basis matrix generator method instead just the basis method. Let's discuss in the weekly.

spharpy/spherical.py
mberz: We agreed to keep it for now and move on to the new sph_harm function which now supports returning the full matrix in the future

tests/test_spherical_harmonics.py
```diff
@@ -8,27 +8,131 @@
...
+                       match="Condon_shortley has to be a bool, or 'auto'."):
+        sh.spherical_harmonic_basis_real(n_max, coords,
+                                         condon_shortley='xx')
+
+
+@pytest.mark.parametrize("implementation", ['spharpy', 'pyfar'])
+def test_spherical_harmonic_default_n10(make_coordinates, implementation):
+    """test the default parameters of SH basis function generator. This
+       simultaneously tests if the methods still match the implementation
+       up to spharpy 0.3.1."""
```
mberz, line 134:
```suggestion
       up to spharpy 0.6.2."""
```

tests/test_spherical_harmonics.py
```diff
@@ -40,13 +144,17 @@ def test_spherical_harmonic_n10(make_coordinates, implementation):
...
     np.testing.assert_allclose(Y, basis, atol=1e-13)
 
 
 @pytest.mark.parametrize("implementation", ['spharpy', 'pyfar'])
-def test_spherical_harmonics_real(make_coordinates, implementation):
+def test_spherical_harmonics_real_n10_default(make_coordinates,
+                                              implementation):
+    """test the default parameters of SH basis function generator. This
+       simultaneously tests if the methods still match the implementation
+       up to spharpy 0.3.1."""
```
mberz, line 157:
```suggestion
       up to spharpy 0.6.2."""
```

spharpy/spherical.py
```diff
@@ -248,17 +249,29 @@ def n3d_to_sn3d_norm(n):
...
+    where:
+
+    - :math:`n` is the degree
+    - :math:`m` is the order
+    - :math:`P_{nm}` is the associated Legendre function
+    - :math:`N_{nm}` is the normalization term
+    - :math:`CS_m` is the Condon-Shortley phase term
+    - :math:`\theta` is the colatitude (angle from the positive z-axis)
+    - :math:`\phi` is the azimuth (angle from the positive x-axis in the
+      xy-plane)
```
mberz, line 274: maybe we can add a link to the pyfar coordinates page here
https://pyfar.readthedocs.io/en/stable/classes/pyfar.coordinates.html (should also work using the intersphinx mapping I guess)

spharpy/spherical.py
mberz: I don't mind both implementations. The current one would be a bit more consistent. 
(The current one is what we agreed on in the last meeting)

f-brinkmann: We decided to leave everything as it is.

spharpy/spherical.py
```diff
@@ -248,32 +248,56 @@ def n3d_to_sn3d_norm(n):
...
     Parameters
     ----------
     n_max : integer
         Spherical harmonic order
-    coordinates : :doc:`pf.Coordinates <pyfar:classes/pyfar.coordinates>`
-        Coordinate object with sampling points for which the basis matrix is
-        calculated
+    coordinates : :py:class:`pf.Coordinates <pyfar.coordinates.Coordinates>` \
+        or :py:class:`sp.SamplingSphere \
+        <spharpy.samplings.coordinates.SamplingSphere>`
```
f-brinkmann, line 287: This does not look right in the docs and the links are not working

spharpy/spherical.py
```diff
@@ -339,9 +385,23 @@ def spherical_harmonic_basis_gradient(n_max, coords):
     ----------
     n_max : int
         Spherical harmonic order
-    coordinates : :doc:`pf.Coordinates <pyfar:classes/pyfar.coordinates>`
-        Coordinate object with sampling points for which the basis matrix is
+    coordinates : :py:class:`pf.Coordinates <pyfar.coordinates.Coordinates>` \
+        or :py:class:`sp.SamplingSphere \
```
f-brinkmann, line 389: See above

spharpy/spherical.py
```diff
@@ -396,21 +488,27 @@ def spherical_harmonic_basis_real(n_max, coords):
...
     Parameters
     ----------
-    n : int
+    n_max : int
         Spherical harmonic order
-    coordinates : :doc:`pf.Coordinates <pyfar:classes/pyfar.coordinates>`
-        Coordinate object with sampling points for which the basis matrix is
+    coordinates : :py:class:`pf.Coordinates <pyfar.coordinates.Coordinates>` \
+        or :py:class:`sp.SamplingSphere \
+        <spharpy.samplings.coordinates.SamplingSphere>`
```
f-brinkmann, line 497: See above

spharpy/spherical.py
```diff
@@ -471,9 +597,23 @@ def spherical_harmonic_basis_gradient_real(n_max, coords):
     ----------
     n_max : int
         Spherical harmonic order
-    coordinates : :doc:`pf.Coordinates <pyfar:classes/pyfar.coordinates>`
-        Coordinate object with sampling points for which the basis matrix is
+    coordinates : :py:class:`pf.Coordinates <pyfar.coordinates.Coordinates>` \
+        or :py:class:`sp.SamplingSphere \
+        <spharpy.samplings.coordinates.SamplingSphere>`
```
f-brinkmann, line 602: See above

f-brinkmann (changes requested):
I made to pushes to fix the formating of code in the docstring and flake8. Code is correctly displayed, wenn I biuld the docs locally, but it does not work on readthedocs. Might this be related to the failing test for building the docs? Is this an issue related to this pull? The test also fails on develop. Maybe fix it there an than merge develop here to see if this resolves the problem.

PR #107, opened 2025-03-07
cookiecutter: fix np.array to scalar deprecation warning

hoyer-a:
fix for

`DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)`


ahms5 (approved):
thank you

tluebeck (approved):
thanks

PR #113, opened 2025-04-30
Dev sh conversion methods

tluebeck:

Make _renormalize and _change_channel_convention from SH signal class (#99) public methods.

### DIFF
--- a/spharpy/spherical.py
+++ b/spharpy/spherical.py
@@ -248,31 +248,35 @@ def n3d_to_sn3d_norm(n):
     return 1 / np.sqrt(2 * n + 1)
 
 
-def renormalize(data, channel_convention, current_norm, target_norm, axis=-2):
+def renormalize(data, channel_convention, current_norm, target_norm, axis):
     """
-    Renormalize either an array of SH coefficients or SH basis functions.
+    Renormalize either an array of spherical harmonics coefficients or basis
+    functions.
 
     Parameters
     ----------
     data : ndarray
-        Data which should be renormalized, either SH
-        coefficients or SH bases
+        Data which should be renormalized, either spherical harmonics
+        coefficients or basis functions.
     channel_convention : str
-        Channel convention of the data which should be renormalized
-    current : str
-        Current channel convention
-    target : str
-        Desired channel convention
+        Channel convention of the data which should be renormalized. Valid
+        conventions are `"acn"` or `"fuma"`.
+    current_norm : str
+        Current normalization. Valid normalizations are `"n3d"`, `"maxN"`, or
+        `"sn3d"`.
+    target_norm : str
+        Desired normalization. Valid normalizations are `"n3d"`,
+        `"maxN"`, or `"sn3d"`.
     axis : integer
-        Axis along which the channel convention should be changed
+        Axis along which the renormalization should be applied
 
     Returns
     -------
-    data : float or complex, ndarray
+    data : ndarray
         Renormalized data
     """
     if target_norm not in ["n3d", "maxN", "sn3d"]:
-        raise ValueError("Invalid normalization, has to be 'sn3d', "
+        raise ValueError("Invalid normalization. Has to be 'sn3d', "
                          f"'n3d', or 'maxN', but is {target_norm}")
     acn = np.arange(data.shape[axis])
 
@@ -305,34 +309,34 @@ def renormalize(data, channel_convention, current_norm, target_norm, axis=-2):
     return data
 
 
-def change_channel_convention(data, current, target, axis=-2):
+def change_channel_convention(data, current, target, axis):
     """
-    Change the channel convention of either an array of SH coefficients, or
-    SH basis functions.
+    Change the channel convention of spherical harmonics coefficients
+    or basis functions.
 
     Parameters
     ----------
     data : ndarray
-        Data of which channel convention should be changed, either SH
-        coefficients or SH bases
+        Data of which channel convention should be changed. Either spherical
+        harmonics coefficients or bases functions.
     current : str
-        Current channel convention
+        Current channel convention. Valid conventions are `"acn"` or `"fuma"`.
     target : str
-        Desired channel convention
+        Desired channel convention. Valid conventions are `"acn"` or `"fuma"`.
     axis : integer
         Axis along which the channel convention should be changed
 
     Returns
     -------
-    data : float or complex, ndarray
+    data : ndarray
         Data with changed channel convention
     """
     if current not in ["acn", "fuma"]:
-        raise ValueError("Invalid channel convention, has to be 'acn', "
+        raise ValueError("Invalid channel convention. Has to be 'acn', "
                          f"or 'fuma', but is {current}")
 
     if target not in ["acn", "fuma"]:
-        raise ValueError("Invalid target channel convention, has to be 'acn', "
+        raise ValueError("Invalid target channel convention. Has to be 'acn', "
                          f"or 'fuma', but is {target}")
 
     acn = np.arange(data.shape[axis])
--- a/spharpy/spherical.py
+++ b/spharpy/spherical.py
@@ -248,11 +248,11 @@ def n3d_to_sn3d_norm(n):
     return 1 / np.sqrt(2 * n + 1)
 
 
-def renormalize(data, channel_convention, current_norm, target_norm):
-    if target not in ["n3d", "maxN", "sn3d"]:
+def renormalize(data, channel_convention, current_norm, target_norm, axis=-2):
+    if target_norm not in ["n3d", "maxN", "sn3d"]:
         raise ValueError("Invalid normalization, has to be 'sn3d', "
                          f"'n3d', or 'maxN', but is {target_norm}")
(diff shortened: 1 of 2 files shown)


spharpy/spherical.py
```diff
@@ -248,6 +248,104 @@ def n3d_to_sn3d_norm(n):
     return 1 / np.sqrt(2 * n + 1)
 
 
+def renormalize(data, channel_convention, current_norm, target_norm, axis=-2):
+    """
+    Renormalize either an array of SH coefficients or SH basis functions.
```
f-brinkmann, line 253:
```suggestion
    Renormalize SH coefficients or basis functions.
```

spharpy/spherical.py
```diff
@@ -248,6 +248,104 @@ def n3d_to_sn3d_norm(n):
...
+
+    Parameters
+    ----------
+    data : ndarray
+        Data which should be renormalized, either SH
+        coefficients or SH bases
+    channel_convention : str
+        Channel convention of the data which should be renormalized
+    current : str
```
f-brinkmann, line 262:
```suggestion
    current_norm : str
```

spharpy/spherical.py
```diff
@@ -248,6 +248,104 @@ def n3d_to_sn3d_norm(n):
...
+        Current channel convention
+    target : str
```
f-brinkmann, line 264:
```suggestion
    target_norm : str
```

spharpy/spherical.py
```diff
@@ -248,6 +248,104 @@ def n3d_to_sn3d_norm(n):
...
+    """
+    Renormalize either an array of SH coefficients or SH basis functions.
+
+    Parameters
+    ----------
+    data : ndarray
+        Data which should be renormalized, either SH
+        coefficients or SH bases
+    channel_convention : str
+        Channel convention of the data which should be renormalized
```
f-brinkmann, line 261: Here and below, I would explicitly mention the options, since it is now a public function.

spharpy/spherical.py
```diff
@@ -248,6 +248,104 @@ def n3d_to_sn3d_norm(n):
...
+        Axis along which the channel convention should be changed
+
+    Returns
+    -------
+    data : float or complex, ndarray
+        Renormalized data
+    """
+    if target_norm not in ["n3d", "maxN", "sn3d"]:
+        raise ValueError("Invalid normalization, has to be 'sn3d', "
+                         f"'n3d', or 'maxN', but is {target_norm}")
```
f-brinkmann, line 276: Since it is a public funtion now, I would suggest these checks for every string input parameter

spharpy/spherical.py
```diff
@@ -248,6 +248,104 @@ def n3d_to_sn3d_norm(n):
...
+        data /= n3d_to_maxn(acn).reshape(shape)
+        if target_norm == "sn3d":
+            data *= n3d_to_sn3d_norm(orders).reshape(shape)
+    return data
+
+
+def change_channel_convention(data, current, target, axis=-2):
+    """
+    Change the channel convention of either an array of SH coefficients, or
+    SH basis functions.
```
f-brinkmann, line 311:
```suggestion
    Change the channel convention of SH coefficients or basis functions.
```

spharpy/spherical.py
```diff
@@ -248,6 +248,104 @@ def n3d_to_sn3d_norm(n):
...
+
+    Parameters
+    ----------
+    data : ndarray
+        Data of which channel convention should be changed, either SH
```
f-brinkmann, line 316:
```suggestion
        Data of which channel convention should be changed. Either SH
```

spharpy/spherical.py
```diff
@@ -248,6 +248,104 @@ def n3d_to_sn3d_norm(n):
...
+        coefficients or SH bases
+    current : str
+        Current channel convention
+    target : str
+        Desired channel convention
```
f-brinkmann, line 321: As above I would suggest to explicitly list the options

spharpy/spherical.py
```diff
@@ -248,6 +248,104 @@ def n3d_to_sn3d_norm(n):
...
+    axis : integer
+        Axis along which the channel convention should be changed
+
+    Returns
+    -------
+    data : float or complex, ndarray
```
f-brinkmann, line 271: I would not mention the dtype since it does not change with respect to the input data
```suggestion
    data : ndarray
```

spharpy/spherical.py
```diff
@@ -248,6 +248,104 @@ def n3d_to_sn3d_norm(n):
...
```
f-brinkmann, line 327: see above
```suggestion
    data : ndarray
```

f-brinkmann (changes requested):
Thanks for bringing this here. Reviewed the functions for now and will check the tests in the next round

spharpy/spherical.py
```diff
@@ -248,6 +248,104 @@ def n3d_to_sn3d_norm(n):
...
+        Renormalized data
+    """
+    if target_norm not in ["n3d", "maxN", "sn3d"]:
+        raise ValueError("Invalid normalization, has to be 'sn3d', "
```
mberz, line 275:
```suggestion
        raise ValueError("Invalid normalization. Has to be 'sn3d', "
```

spharpy/spherical.py
mberz: Since it's not mentioned anywhere. Please spell out SH everywhere for improved readability

spharpy/spherical.py
```diff
@@ -248,6 +248,104 @@ def n3d_to_sn3d_norm(n):
     return 1 / np.sqrt(2 * n + 1)
 
 
+def renormalize(data, channel_convention, current_norm, target_norm, axis=-2):
```
mberz, line 251:
```suggestion
def renormalize(data, channel_convention, current_norm, target_norm, axis):
```
Using -2 as default axis does not seem like the best option here. For example for a basis matrix it is -1.
So I would prefer forcing users to choose the appropriate axis.

spharpy/spherical.py
```diff
@@ -248,6 +248,104 @@ def n3d_to_sn3d_norm(n):
...
+
+    if current_norm == 'maxN':
+        # convert to n3d
+        data /= n3d_to_maxn(acn).reshape(shape)
+        if target_norm == "sn3d":
+            data *= n3d_to_sn3d_norm(orders).reshape(shape)
+    return data
+
+
+def change_channel_convention(data, current, target, axis=-2):
```
mberz, line 308:
```suggestion
def change_channel_convention(data, current, target, axis):
```
Same comment as above

mberz (changes requested):
I would prefer removing the default option for the axis argument.

Otherwise only minor docstring comments

spharpy/spherical.py
```diff
@@ -248,6 +248,108 @@ def n3d_to_sn3d_norm(n):
...
+        `"maxN"`, or `"sn3d"`.
+    axis : integer
+        Axis along which the renormalization should be applied
+
+    Returns
+    -------
+    data : ndarray
+        Renormalized data
+    """
+    if target_norm not in ["n3d", "maxN", "sn3d"]:
```
f-brinkmann, line 278: I would suggest to make this check also for `current_norm` and `channel_convention`

tests/test_spherical.py
```diff
@@ -0,0 +1,65 @@
...
+    sh_data = np.ones((4, 2))
+    current_norm = 'n3d'
+
+    # test invalid type
+    with raises(ValueError, match="Invalid normalization. Has to be 'sn3d', "
+                                  "'n3d', or 'maxN', but is wrong_norm"):
+        sh.renormalize(sh_data, 'acn', current_norm, 'wrong_norm', axis=0)
+
+    # test maxN
+    sh_data_renorm = sh.renormalize(sh_data.copy(), 'acn', current_norm,
```
f-brinkmann, line 19: It's maybe a little dangerous to require the user to copy the data if it should now be overwritten. In line with other pyfar functions, I would suggest to take the copy inside the functions
```suggestion
    sh_data_renorm = sh.renormalize(sh_data, 'acn', current_norm,
```

tests/test_spherical.py
```diff
@@ -0,0 +1,65 @@
...
+    np.testing.assert_equal(sh_data_renorm,
+                            sh_data_ref)
+    # test back to n3d
+    sh_data_renorm = sh.renormalize(sh_data.copy(), 'acn', current_norm, 'n3d',
+                                    axis=-2)
+    np.testing.assert_equal(sh_data_renorm,
+                            np.ones((4, 2)))
+
+
+def test_change_channel_convention():
```
f-brinkmann, line 46: I would suggest to add `test_change_chanel_convention_errors` for testing if all errors are raised as expected

f-brinkmann (changes requested):
I ran coverage and there are untested but important lines in `renormalize`

284, 300-302, 306-308. Can you please add tests for those as well?

f-brinkmann (approved):
Thanks, looks good to me now :)

mberz (approved):

PR #114, opened 2025-05-06
DOC: remove wrong line in doc

ahms5:
there is a line, which is not correct in the docstring of spherical_harmonic_basis.
this is fixed now.


hoyer-a (approved):
Thanks for catching

f-brinkmann (approved):

mberz (approved):

PR #115, opened 2025-05-06
fix intershinx links

ahms5:

- the intershinx links to pyfar are fixed now and should create proper links in the doc
- link is also set for spharpy.SamplingSphere, but not woking yet, because it is not part of the doc yet

### DIFF
--- a/spharpy/plot/spatial.py
+++ b/spharpy/plot/spatial.py
@@ -48,7 +48,7 @@ def scatter(coordinates, ax=None):
 
     Parameters
     ----------
-    coordinates : :doc:`pf.Coordinates <pyfar:classes/pyfar.coordinates>`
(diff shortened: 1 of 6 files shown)


spharpy/plot/spatial.py
```diff
@@ -62,7 +62,7 @@ def scatter(coordinates, ax=None):
         >>> spharpy.plot.scatter(coords)
 
 
-    """
+    """  # noqa: E501
```
xefonon, line 65: why is ruff complaining here?

f-brinkmann (approved):
Thanks

tluebeck (dismissed):
Thanks, looks good to me.

xefonon (approved):
Nice! :)

mberz:
Ignoring dimissal of reviews after removing a obsolete whitespace in docstring.

PR #117, opened 2025-05-12
fix matplotlib version

### LINKED ISSUE #91: Contour map plot fails

f-brinkmann:

Closes #91 

Force matplotlib version >= 3.10.3 to avoid error using contour map plots. Unfortunately it seems that this does not work vor Python 3.9

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,13 +23,13 @@ classifiers = [
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
     "Programming Language :: Python :: 3.13",
-    
+
 ]
 dependencies = [
     'numpy>=1.22',
     'scipy',
     'urllib3',
-    'matplotlib>=3.3.0',
+    'matplotlib>=3.10.3',
     'pyfar<0.8.0',
     'pytz',
 ]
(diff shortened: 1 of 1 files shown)


ahms5:
I think we can also start deprecating 3.9 where needed. numpy also did it since [2.1.0](https://numpy.org/doc/stable/release/2.1.0-notes.html)

mberz:
#123  and #121 will not be merged into main, so requiring pyfar>0.7.3 is not required.
I would prefer not over-constraining if not necessary.

f-brinkmann:
They will be with spharpy 1.0.0. In this case we need a sparate pull to upgrade to pyfar v0.7.3 that we merge before the upcoming major spharpy release

mberz (approved):

hoyer-a (approved):

Issue #119, opened 2025-05-16
Additional properties to SamplingSphere


mberz:
- Enforce a unique radius (under some numerical precision)
- Kind of sampling (quadrature boolean)

Changes potentially affect #101

f-brinkmann:
@ahms, @xefonon, @mberz Elegantly enforcing a unique radius seems to be non-trivial. The radius must be checked during initialization and after calling any coordinate setter, e.g., after `sampling_sphere.elevation = 0`. The problem is that the latter is done in `Coordinates` not `SamplingSphere`.

Possible solutions
- override all coordinates setters in `SamplingSphere`. This would be prone to errors, because it would require changes in case new coordinates are introduced in `Coordinates`.
- check only upon initialization. This would not be very strict.
- modify `Coordinates` e.g. by adding a function such as `_check_upon_coordinates_change()` that is called each time a coordinate changes. This could not be done right away, because it requires changes in pyfar.

The quadrature property has a similar problem. If the weights of a Coordinates object are changed to `None` this would require to trigger a `sampling_sphere.quadrature = False` action. Here a workaround might be to define only a setter

```python
@property
def quadrature(self):
    # there could be other checks here as well, e.g., if all weights are >0 and sum to 1 (4 pi?)
    return False if self.weights is None else True
```

ahms5:
internally the ``self._set_points(x, y, z)`` is always called if sth is set/init, independend of the coordinate system. it is the only method which writes to Coordinate._x Coordinate._y and Coordinate._z.
maybe we could overwrite this method and add a radius check at the beginning before calling the parent class.
we can simple calculate the radius from cartesian coords.

ahms5:
it is merged, so we can close it for now.

Issue #77, opened 2024-10-18
Ceck consistency of weights


f-brinkmann:
Check if weights normalize to 1 in SamplingSphere as in pyfar Coordinates objects.

mberz:
They should normalize to 4*pi, not 1.

ahms5:
and in pyfar.Coordiantes the sum of the weight is not checked at all. It left upon the user.

mberz:
In #123 I discovered that when moving the samplings to pyfar and back, the weights were modified and now sum to 1.

ahms5:
it is merged, so we can close it for now.

PR #120, opened 2025-05-16
Feature quadrature property for SamplingSphere

f-brinkmann:

Part of #77 and #119.

This would be a possible way to implement a `quadrature` property (getter only). If this gets approved I would

- update all samplings to make sure weights sum to 4 pi
- for all samplings that are quadratures, add a test if `quadrature == True`
- add and update docstrings
- Strictly enforce using SamplingSphere over Coordinates for spherical samplings and tests thereof


xefonon:
Hey @f-brinkmann, this isn't finished right? Also I should wait for this to be merged before finishing #101 right?

f-brinkmann:
Yes, not finished. This is only the bare code. If we agree that this is what we want, I'll add docstring, testing, and modification of the weights to sum to 4 pi. I guess, you depened on this now to finish the SH class. Lets hope for quick comments...

xefonon:
> Yes, not finished. This is only the bare code. If we agree that this is what we want, I'll add docstring, testing, and modification of the weights to sum to 4 pi. I guess, you depened on this now to finish the SH class. Lets hope for quick comments...

I think we should go ahead with this, I can start with some comments

f-brinkmann:
Sure - I think you can also merge this into your branch, already and update later. You only depend on the fact that the property exists, but not how it is implemented in detail.

spharpy/samplings/coordinates.py
```diff
@@ -310,3 +310,17 @@ def n_max(self, value):
             self._n_max = None
         else:
             self._n_max = int(value)
+
```
xefonon, line 313: Not sure `n_max` setter/getter is used anywhere. Is it worth keeping in this class?

f-brinkmann: I think its good to inform the user about the theoretical max order that can be used with a gird. We could think about using it in the SH class, but I guess that would be something for discussion during the weekly.

spharpy/samplings/coordinates.py
```diff
@@ -310,3 +310,17 @@ def n_max(self, value):
...
+    @property
+    def quadrature(self):
```
ahms5, line 315: from what I understood last time the quadrature should be set during intitialization e.g. in the samplings module. If we derive it from the weights it does not contain any extra information. If I add weights later by voronoi, it will result in wrong result here.

ahms5 (changes requested):
thank you for taking care, not sure I misunderstand anything last time.

spharpy/samplings/coordinates.py
ahms5: the SamplingSphere class is just a conatiner for SamplingSpheres, so for every grid with correct weight it should sum up to 4pi, independed of beeing a quadrature or not. The SamplingSphere is for example generated in samplings module, where we know each time if it is a quadrature or not. it does not make sence in my optionon to guess later on if it is quadrature or not, especially if we know it for sure at a earlier stage. This can cause much more errors, becuase users will assume that the property is giving the right result. 
Also Users should just use the class if they have a samplingSphere and not any coordinates, so they should know if their sampling is a quadrature or not. we could use the default as False, then nothing will go wrong in using it in the SH class with defaults

f-brinkmann: My understanding was that these are exactly the requirements for numerical integration (as used in the inverse transform), no matter if the weights are analytically computed or using Voronoi tesseleation.

mberz: I don't think that implementing it like this is the best idea. Quadrature should only be true if the grid is designed to match a feasible quadrature rule. In other cases the weights can be used for weighted least-squares.
See https://pub.dega-akustik.de/NAG_DAGA_2009/data/articles/000296.pdf

spharpy/samplings/samplings.py
```diff
@@ -153,14 +153,15 @@ def hyperinterpolation(n_points=None, n_max=None, radius=1.):
         sep=' ').reshape((int(n_points), 4))
 
     # normalize weights
```
ahms5, line 155:
```suggestion
    # normalize weights to 4pi
```
maybe also add in the docstring that the weights are normalized to 4pi

ahms5: and fix the doc in general SamplingSphere is returned and we can also mention about weather its a quadrature or not

spharpy/samplings/samplings.py
```diff
@@ -487,12 +490,12 @@ def equiangular(n_points=None, n_max=None, radius=1.):
 
     # repeat and normalize sampling weights
     w = np.tile(w, n_phi)
-    w = w / np.sum(w)
```
ahms5, line 490: same here, fix documentation output to samplingSphere and tell about normalization of the sphere weights

spharpy/samplings/coordinates.py
```diff
@@ -30,21 +30,32 @@ def __init__(
...
         comment : str, optional
             Comment about the stored coordinate points. The default is
             ``""``, which initializes an empty string.
         sh_order : int, optional
             Maximum spherical harmonic order of the sampling grid.
             The default is ``None``.
+        quadrature : bool, optional
+            Flag that indicates if points belong to a quadrature, which
+            requires that all `weights` are greater than zero and sum to
+            :math:`4\pi`. The default is ``False``.
```
mberz, line 46:
```suggestion
```
Duplicate

spharpy/samplings/coordinates.py
```diff
@@ -310,3 +348,33 @@ def n_max(self, value):
...
+        # check input
+        if not isinstance(value, bool):
+            raise TypeError(
+                f'quadrature must be True or False but is {value}')
+
+        # flag indicating if weights sum to 4 pi
+        if self.weights is None:
+            weights_sum = False
+        else:
+            weights_sum = np.abs(np.sum(self.weights) - 4 * np.pi) < \
```
ahms5, line 370: not sure if this should be tested here or in the sh class, I guess @mberz can clarify

spharpy/samplings/coordinates.py
```diff
@@ -30,21 +30,32 @@ def __init__(
...
+            Flag that indicates if points belong to a quadrature, which
+            requires that all `weights` are greater than zero and sum to
+            :math:`4\pi`. The default is ``False``.
         """
         pf.Coordinates.__init__(
             self, x, y, z, weights=weights, comment=comment)
         self._n_max = n_max
 
+        self._quadrature = None
+        self.quadrature = quadrature
```
mberz, line 53: Why is this first set to None and then overwritten?

ahms5 (changes requested):
Thank you, just small comments about the docs, especially in samplings, we could inpruve the return parater description here, somethimes it is even wrong which type is really returned

spharpy/samplings/coordinates.py
```diff
@@ -310,3 +348,33 @@ def n_max(self, value):
...
+            weights_sum = False
+        else:
+            weights_sum = np.abs(np.sum(self.weights) - 4 * np.pi) < \
+                np.finfo(self.weights.dtype).resolution * 5
+
+        # check requirements
+        if value and (np.any(self.weights <= 0) or not weights_sum):
+            raise ValueError(
+                'quadrature can not be True because at least one weight is '
+                'not greater than zero and/or weights do not sum to 4 pi')
```
mberz, line 377: After merging #123 this can be simplified to only check if weights are not None

mberz (changes requested):

spharpy/samplings/coordinates.py
mberz: see #123

spharpy/samplings/samplings.py
```diff
@@ -487,12 +490,12 @@ def equiangular(n_points=None, n_max=None, radius=1.):
...
+    w = w / np.sum(w) * 4 * np.pi
```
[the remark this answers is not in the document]
f-brinkmann: The weights sum up to 19.63 instead of 4 pi without the normalization for the test in test_samplings.py. There might be an error in our implementation that should be fixed in a different pr.

spharpy/samplings/samplings.py
```diff
@@ -572,12 +575,12 @@ def gaussian(n_points=None, n_max=None, radius=1.):
 
     # compute the sampling weights
     weights = np.tile(weights, n_phi)
-    weights = weights / np.sum(weights)
+    weights = weights / np.sum(weights) * 4 * np.pi
```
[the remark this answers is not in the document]
f-brinkmann: The weights sum up to 10 instead of 4 pi without the normalization for the test in test_samplings.py. There might be an error in our implementation that should be fixed in a different pr.

spharpy/samplings/samplings.py
```diff
@@ -1067,14 +1071,14 @@ def lebedev(n_points=None, n_max=None, radius=1.):
     leb = lebedev_sphere(n_points)
 
     # normalize the weights
-    weights = leb["w"] / (4 * np.pi)
+    weights = leb["w"] / np.sum(leb["w"]) * 4 * np.pi
```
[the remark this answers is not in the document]
f-brinkmann: corrected

spharpy/samplings/samplings.py
f-brinkmann: added a note to the docstring of all quadrature samplings.

spharpy/samplings/samplings.py
```diff
@@ -1235,12 +1239,15 @@ def fliege(n_points=None, n_max=None, radius=1.):
         variable_names=f"Fliege_{int(n_points)}")
     fliege = fliege[f"Fliege_{int(n_points)}"]
 
+    # normlize weights
+    weights = fliege[:, 2] / np.sum(fliege[:, 2]) * 4 * np.pi
```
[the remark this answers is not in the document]
f-brinkmann: corrected

spharpy/samplings/coordinates.py
```diff
@@ -125,13 +145,14 @@ def from_spherical_elevation(
         """
 
         x, y, z = sph2cart(azimuth, np.pi / 2 - elevation, radius)
-        return cls(
-            x, y, z, weights=weights, comment=comment, n_max=n_max)
+        return cls(x, y, z, weights=weights, comment=comment, n_max=n_max,
+                   quadrature=quadrature)
```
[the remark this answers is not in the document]
f-brinkmann: reverted everywhere

mberz (changes requested):
I don't think hyperinterpolation (or ciritical sampling) is a quadrature grid. Please change that (or correct me if I'm wrong)
Otherwise a couple of small things.

mberz (approved):

PR #121, opened 2025-05-16
Check for unique radius in SamplingSphere

f-brinkmann:

Refers to #119 

- add check for unique radius in SamplingSphere
- add tests thereof

### DIFF
--- a/spharpy/samplings/coordinates.py
+++ b/spharpy/samplings/coordinates.py
@@ -310,3 +310,23 @@ class SamplingSphere(pf.Coordinates):
             self._n_max = None
         else:
             self._n_max = int(value)
+
+    def _set_points(self, x, y, z):
+
+        # cast to numpy array
+        x = np.atleast_1d(np.asarray(x, dtype=np.float64))
+        y = np.atleast_1d(np.asarray(y, dtype=np.float64))
+        z = np.atleast_1d(np.asarray(z, dtype=np.float64))
+
+        # check if all points have the same radius
+        radius = np.sqrt(x.flatten()**2 + y.flatten()**2 + z.flatten()**2)
+        radius_delta = np.max(np.abs(np.mean(radius) - radius))
+        radius_tolerance = np.finfo(x.dtype).resolution * 5
+        if radius_delta > radius_tolerance:
+            raise ValueError(
+                'All points must have the same radius but the deviation from '
+                f'the mean radius is {radius_delta:.3g} m, which exceeds the'
+                f' tolerance of {radius_tolerance:.3g} m.')
+
+        # set points
+        super()._set_points(x, y, z)
--- a/tests/test_coordinates.py
+++ b/tests/test_coordinates.py
@@ -1,4 +1,5 @@
 from spharpy import SamplingSphere
+import pytest
 
 
 def test_sampling_sphere_init():
@@ -35,3 +36,27 @@ def test_setter_n_max():
 
     sampling.n_max = n_max
     assert sampling._n_max == n_max
+
+
+def test_error_multiple_radius_initialization():
+    """
+    Test if entering points with multiple radii during initialization raises
+    an error.
+    """
+
+    match = '0.5 m, which exceeds the tolerance of 5e-15 m'
+    with pytest.raises(ValueError, match=match):
+        SamplingSphere([1, 0], 0, 0)
+
+
+def test_error_multiple_radius_setter():
+    """
+    Test if entering points with multiple radii after initialization raises
+    an error.
+    """
+
+    sampling_sphere = SamplingSphere([1, 1], 0, 0)
+
+    match = '0.5 m, which exceeds the tolerance of 5e-15 m'
+    with pytest.raises(ValueError, match=match):
+        sampling_sphere.x = [1, 0]


spharpy/samplings/coordinates.py
```diff
@@ -310,3 +310,23 @@ def n_max(self, value):
             self._n_max = None
         else:
             self._n_max = int(value)
+
+    def _set_points(self, x, y, z):
```
xefonon, line 314: Do you think there should be input validation? If `x,y,z` are `None` or if they have different shapes for example?

spharpy/samplings/coordinates.py
```diff
@@ -310,3 +310,23 @@ def n_max(self, value):
...
+
+        # cast to numpy array
+        x = np.atleast_1d(np.asarray(x, dtype=np.float64))
+        y = np.atleast_1d(np.asarray(y, dtype=np.float64))
+        z = np.atleast_1d(np.asarray(z, dtype=np.float64))
+
+        # check if all points have the same radius
+        radius = np.sqrt(x.flatten()**2 + y.flatten()**2 + z.flatten()**2)
+        radius_delta = np.max(np.abs(np.mean(radius) - radius))
+        radius_tolerance = np.finfo(x.dtype).resolution * 5
```
xefonon, line 324: why is this multiplied by 5?

xefonon (commented):
Thanks, only have a comment regarding the inputs of the function.

spharpy/samplings/coordinates.py
f-brinkmann: All input checks are done in `Coordinates._set_points`. But its actually a good point. Now this is called first to make sure data is valid or an appropriate error is raised.

spharpy/samplings/coordinates.py
f-brinkmann: More or less arbitrary an arbitrary tolerance. I added a comment to make this clear. Do you think this is fine or to strict/relaxed?

xefonon: Sure, but yes I think it is too strict. This would be something of the order `1e-15` where we would want a tolerance of 1 mm perhaps?

f-brinkmann: 1 mm would be fine for me. Lets wait for other comments until I change this.

spharpy/samplings/coordinates.py
```diff
@@ -310,3 +310,22 @@ def n_max(self, value):
...
+
+    def _set_points(self, x, y, z):
+        # set points first to make sure data is valid
+        super()._set_points(x, y, z)
+
+        # check if all points have the same radius
+        radius = np.sqrt(
+            self.x.flatten()**2 +
+            self.y.flatten()**2 +
+            self.z.flatten()**2)
```
ahms5, line 322:
```suggestion
        radius = self.radius
```
since it is already set here self.radius should work as expected.

ahms5 (dismissed):
thanks, looks fine, just a minor comment

spharpy/samplings/coordinates.py
f-brinkmann: Thanks for catching. In the initial version points were set after testing for the radius but this is more elegant and less prone to errors.

ahms5 (dismissed):

spharpy/samplings/coordinates.py
```diff
@@ -310,3 +310,19 @@ def n_max(self, value):
...
+
+    def _set_points(self, x, y, z):
+        # set points first to make sure data is valid
+        super()._set_points(x, y, z)
+
+        # check if all points have the same radius
+        radius = self.radius
+        radius_delta = np.max(np.abs(np.mean(radius) - radius))
+        # add a tolerance of 1 mm
+        radius_tolerance = .001
```
mberz, line 322: This could be simplified using np.unique(radius) or np.unique(np.round(radius, desired_precision))
https://numpy.org/doc/stable/reference/generated/numpy.unique.html

mberz (commented):
Optional code improvement. Otherwise looking good.

spharpy/samplings/coordinates.py
mberz, line 322: Using a fixed absolute tolerance is probably not the best idea here.
I would suggest making this relative, or even better make this tolerance a property of the class.

f-brinkmann: np.round would only work if we use a decimal precision. But since we used tolerance values in other parts of pyfar, I chose to implement it this way.

spharpy/samplings/coordinates.py
```diff
@@ -36,7 +36,17 @@ def __init__(
...
             The default is ``None``.
+        radius_tolerance : float, optional
+            All points that are stored in a SamplingSphere must have the same
+            radius and an error is raised if the maximum deviation from the
+            mean radius exceeds this tolerance. The default of ``1e-6`` meter
+            is intended to allow for some numerical inaccuracy.
         """
+        # must be initialized first, because it is already required for
+        # checking if input points have the same radius
+        self._radius_tolerance = None
```
ahms5, line 47:
```suggestion
```
is this line required?

spharpy/samplings/coordinates.py
```diff
@@ -8,7 +8,7 @@ class SamplingSphere(pf.Coordinates):
 
     def __init__(
             self, x=None, y=None, z=None, n_max=None, weights: np.array = None,
-            comment: str = ""):
+            comment: str = "", radius_tolerance=1e-6):
```
ahms5, line 11: and now we need to add it for all from* methods, because right now we cannot set it via e.g. ``SamplingSphere.from_spherical_elevation(radius_tolerance=.1)``

this could be also done in an other pr together with quadrature attribute.

spharpy/samplings/coordinates.py
```diff
@@ -36,7 +36,17 @@ def __init__(
         sh_order : int, optional
             Maximum spherical harmonic order of the sampling grid.
             The default is ``None``.
+        radius_tolerance : float, optional
```
ahms5, line 39: I would prefere a relative tolerance, but i would be fine with absolute as well.
we could call it radius_atol or radius_rtol, then its directly clear whch kind of tolerance it is.

ahms5 (changes requested):
Thank you!

spharpy/samplings/coordinates.py
```diff
@@ -310,3 +320,31 @@ def n_max(self, value):
...
+        super()._set_points(x, y, z)
+
+        # check if all points have the same radius
+        radius = self.radius
+        radius_delta = np.max(np.abs(np.mean(radius) - radius))
+        if radius_delta > self.radius_tolerance:
+            raise ValueError(
+                'All points must have the same radius but the deviation from '
+                f'the mean radius is {radius_delta:.3g} m, which exceeds the'
+                f' tolerance of {self.radius_tolerance:.3g} m.')
```
mberz, line 349: This set's the points (in super()._set_points) even if the ValueError is raised.
That's definitely a bug ;)

mberz (changes requested):

spharpy/samplings/coordinates.py
ahms5: why is it a bug?

if so, we could change the internal handling in pyfar.Coordinates, to split the current ``_set_points`` into ``check_points``to to all the checked and braodcasting and then the ``_set_points``
(same for _set_weights)
we could do it for 0.8.0, it just changing the backend, so no need to deprecate anyting in pyfar

mberz: This is how I interpret the code, please correct me if I'm wrong here.
You want to avoid setting points which do not share the same radius, but in the process the points are set anyways. Afterwards an error is raised indicating that the operation failed, however, the internal class values have already been overwritten before the error was raised.

f-brinkmann: I agree with mberz and will switch the order. Check first, setting the data second.

spharpy/samplings/coordinates.py
f-brinkmann: Opted to rename it to radius_atol

spharpy/samplings/coordinates.py
f-brinkmann: Done it directly here

spharpy/samplings/coordinates.py
f-brinkmann: Yes otherwise the setter fails

spharpy/samplings/coordinates.py
```diff
@@ -310,3 +361,35 @@ def n_max(self, value):
...
+        """Get or set the radius tolerance in meter."""
+        if isinstance(value, (int, float)) and value > 0:
+            self._radius_atol = float(value)
+        else:
+            raise ValueError(
+                'The radius tolerance must be a number greater than zero')
+
+    def _set_points(self, x, y, z):
+
+        # check for unique radius
```
ahms5, line 381: instead of copy and pasting the code from pyfar we could just use the _check_points method here before checking the radius. this will do all common checks without setting the results. this will prevent the bug indicated by @mberz and we wouldnt have duplicated code

https://github.com/pyfar/pyfar/pull/803

spharpy/samplings/coordinates.py
f-brinkmann: This is now handeled in super._check_points()

spharpy/samplings/coordinates.py
f-brinkmann: done

spharpy/samplings/coordinates.py
f-brinkmann: we have absolute tolerances in others places in pyfar, so I would prefer to leave it like this. Since its intended to account for numerical noise, I guess it is fine as well.

spharpy/samplings/coordinates.py
```diff
@@ -303,3 +354,36 @@ def n_max(self, value):
             self._n_max = None
         else:
             self._n_max = int(value)
+
+    @property
+    def radius_atol(self):
```
mberz, line 359: I find the name of the property a bit confusing. How about renaming it to simply `radius_tolerance` ?
There is not rtol, which would make the distinction necessary and the docstring also does not really help to clarify what atol means.

mberz (changes requested):

spharpy/samplings/coordinates.py
f-brinkmann: agreed and changed

mberz (approved):

spharpy/samplings/coordinates.py
```diff
@@ -204,14 +241,16 @@ def from_spherical_side(
...
         return cls(
-            x, y, z, weights=weights, comment=comment, n_max=n_max)
+            x, y, z, weights=weights, comment=comment, n_max=n_max,
+            radius_tolerance=radius_tolerance)
 
     @classmethod
     def from_spherical_front(
             cls, frontal, upper, radius, n_max=None, weights: np.array = None,
-            comment: str = ""):
+            comment: str = "", radius_tolerance: float = 1e-6):
```
xefonon, line 253: I only have one comment: I think this default value is too strict. _Robots_ have a tolerance of 1e-5 for positioning in Cartesian coordinates. I doubt us human beings will be this accurate.

f-brinkmann: Would it be fine to keep the tolerance as it is, but clearly to point to the parameter in the corresponding error message? I think this would make people aware of the idea behind this and make them think about a reasonable tolerance for their application.

spharpy/samplings/coordinates.py
mberz: sorry, I did not see the comment when reviewing and suggested to not name it atol. 
Still, I think atol is not as intuitive as simply naming it tolerance.

spharpy/samplings/coordinates.py
```diff
@@ -121,14 +143,17 @@ def from_spherical_elevation(
         >>> sampling = pf.SamplingSphere.from_spherical_elevation(0, 0, 1)
         """
 
-        x, y, z = sph2cart(azimuth, np.pi / 2 - elevation, radius)
+        x, y, z = sph2cart(
+            azimuth, np.pi / 2 - np.atleast_1d(elevation), radius)
```
ahms5, line 147: what is this doing here?

spharpy/samplings/coordinates.py
ahms5: is this still the case? I cant find a reason, why this should happen.
I think we should avoid setting the private variable out of the setter.

spharpy/samplings/coordinates.py
```diff
@@ -303,3 +354,36 @@ def n_max(self, value):
...
+
+    def _check_points(self, x, y, z):
+        """Check input data before setting coordinates"""
+
+        # convert to numpy arrays of the same shape
+        x, y, z = super()._check_points(x, y, z)
+
+        # check for equal radius
+        radius = np.sqrt(x.flatten()**2 + y.flatten()**2 + z.flatten()**2)
+        radius_delta = np.max(np.abs(np.mean(radius) - radius))
```
ahms5, line 380: should we check here, between the largest and the smallest radius?

I found your example in the test with 0,0,0 and 1,0,0 and the error is returning 0.5m which does not really mean anything.
also using mean or median, could lead to different behaviour here. The distance between the lowest and largest radius is more meaningful in my point of view.

I would suggest the largest distance beween the radius, so max-min

spharpy/samplings/coordinates.py
```diff
@@ -303,3 +354,36 @@ def n_max(self, value):
...
+
+    @property
+    def radius_tolerance(self):
+        """Get or set the radius tolerance in meter."""
+        return self._radius_tolerance
+
+    @radius_tolerance.setter
+    def radius_tolerance(self, value):
+        """Get or set the radius tolerance in meter."""
+        if isinstance(value, (int, float)) and value > 0:
```
ahms5, line 366: is we change the setter on runtime, we need to check the radius again.
maybe its better to set the radius at the beginning and dont allow to change it anymore. otherwise we could call 

self._x,  self._y,  self._z = self._check_points(self._x,  self._y,  self._z)

ahms5 (commented):
Thank you for taking care!

spharpy/samplings/coordinates.py
f-brinkmann: I did this as suggested

PR #123, opened 2025-05-20
Check sum of weights in SamplingSphere

### LINKED ISSUE #77: Ceck consistency of weights

mberz:

Closes #77 

- [x] Add a check if weights used in SamplingSphere sum to the surface integral on the unit-sphere

Weights will be fixed in #120 to avoid conflicts.

### DIFF
--- a/spharpy/samplings/coordinates.py
+++ b/spharpy/samplings/coordinates.py
@@ -41,9 +41,9 @@ class SamplingSphere(pf.Coordinates):
             self, x, y, z, weights=weights, comment=comment)
         self._n_max = n_max
 
-        self.weights.__doc__ = (
-            "The area or quadrature weights of the sampling. "
-            "Their sum must be equal to 4*pi.")
+        # self.weights.__doc__ = (
+        #     "The area or quadrature weights of the sampling. "
+        #     "Their sum must be equal to 4*pi.")
 
     @classmethod
     def from_cartesian(
@@ -315,28 +315,14 @@ class SamplingSphere(pf.Coordinates):
         else:
             self._n_max = int(value)
 
-    def _set_weights(self, weights):
-        """Check and set sampling weights.
+    def _check_weights(self, weights):
+        weights = super()._check_weights(weights)
 
-        Set self._weights, which is an atleast_1d numpy array of shape
-        [L,M,...,N].
-
-        """
-        # check if weights is None
         if weights is None:
-            self._weights = weights
-            return
-
-        weights = np.asarray(weights, dtype=np.float64)
-
-        # reshape according to self._points
-        if weights.size != self.csize:
-            raise ValueError("Weights must have same size as csize.")
-        weights = weights.reshape(self.cshape)
-
-        if not np.isclose(np.sum(weights), 4*np.pi, atol=1e-6, rtol=1e-6):
+            return weights
+        elif not np.isclose(np.sum(weights), 4*np.pi, atol=1e-6, rtol=1e-6):
             raise ValueError(
                 "The sum of the weights must be equal to 4*pi. "
                 f"Current sum: {np.sum(weights)}")
 
-        self._weights = weights
+        return weights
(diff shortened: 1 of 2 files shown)


spharpy/samplings/coordinates.py
```diff
@@ -310,3 +310,29 @@ def n_max(self, value):
...
             self._n_max = int(value)
+
+    def _set_weights(self, weights):
+        """Check and set sampling weights.
+
+        Set self._weights, which is an atleast_1d numpy array of shape
+        [L,M,...,N].
+
+        """
+        # check if weights is None
```
ahms5, line 321: any special reason why not to all the _weights from Coordinates first?
I mean it setting it before the sum is checked, but its also raising an error so it does not really matter.
https://github.com/pyfar/pyfar/blob/02e8695b714e3ce380522e423c174284c6137020/pyfar/classes/coordinates.py#L2734

mberz: Same reason as in #121 
If super().method is used, the incorrect weights can be set in the process, which is to be avoided.

spharpy/samplings/coordinates.py
```diff
@@ -310,3 +314,29 @@ def n_max(self, value):
...
+            return
+
+        weights = np.asarray(weights, dtype=np.float64)
+
+        # reshape according to self._points
+        if weights.size != self.csize:
+            raise ValueError("Weights must have same size as csize.")
+        weights = weights.reshape(self.cshape)
+
+        if not np.isclose(np.sum(weights), 4*np.pi, atol=1e-6, rtol=1e-6):
```
xefonon, line 337: You don’t check for negativity or nans. Unnormalized or negative weights could silently pass the sum‐check yet break algorithms downstream.

spharpy/samplings/coordinates.py
```diff
@@ -310,3 +314,29 @@ def n_max(self, value):
...
+            self._weights = weights
+            return
+
+        weights = np.asarray(weights, dtype=np.float64)
+
+        # reshape according to self._points
+        if weights.size != self.csize:
+            raise ValueError("Weights must have same size as csize.")
+        weights = weights.reshape(self.cshape)
+
```
xefonon, line 336:
```suggestion
sigma_w = np.sum(weights)
```
To use in place of np.sum(weights) which is called twice.

spharpy/samplings/coordinates.py
```diff
@@ -310,3 +314,29 @@ def n_max(self, value):
...
+        Set self._weights, which is an atleast_1d numpy array of shape
+        [L,M,...,N].
+
+        """
+        # check if weights is None
+        if weights is None:
+            self._weights = weights
+            return
+
```
xefonon, line 329:
```suggestion
    if not isinstance(weights, (collections.abc.Sequence, np.ndarray)):
        raise TypeError("Weights must be array-like, not a %r" % type(weights))
```
If you pass an object array whose elements are all numeric, np will quietly convert them to floats.

xefonon (commented):
Thanks @mberz I had a few micro (nano)-comments

spharpy/samplings/coordinates.py
mberz: Good point!

f-brinkmann (approved):
Just two minimal comments on the docs, otherwise approved

spharpy/samplings/coordinates.py
```diff
@@ -310,3 +324,32 @@ def n_max(self, value):
             self._n_max = None
         else:
             self._n_max = int(value)
+
+    def _check_weights(self, weights):
+        weights = super()._check_weights(weights)
```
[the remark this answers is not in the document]
mberz: Ah, I think that got lost somewhere when shortening the commit history.

f-brinkmann (approved):

PR #126, opened 2025-05-27
Fix incorrect samplings and respective weights

### LINKED ISSUE #125: Weights do not sum to 4 pi
Weights for the gaussian and equiangular sampling do not sum to 4 pi but they should. For the Gaussian sampling this seems to be the case generally. For the equiangular sampling this only happens if specifiying `n_points` instead of `n_max`.

mberz:

Closes #125 and related to #77 

All of these issues are only present on develop. Main is not affected.

- Fixed incorrect design of Gaussian sampling when using n_points instead of n_max
- Recovered old weight calculation in Gaussian quadrature
- Equiangular designs are only valid for even numbers of points and the same equiangle steps in azimuth and elevation. Set weights and quadrature property appropriately.
- Updated and extended testing

### DIFF
--- a/spharpy/samplings/samplings.py
+++ b/spharpy/samplings/samplings.py
@@ -555,14 +555,15 @@ def gaussian(n_points=None, n_max=None, radius=1.):
         raise ValueError(
             "Either the n_points or n_max needs to be specified.")
 
-    if (
-        n_points is not None and
-        (type(n_points) is not int or (np.asarray(n_points).size > 1))
-    ):
-        raise ValueError(
-            "The number of points needs to be a positive natural number. ",
-            f"Instead it is{n_points}",
-        )
+    elif (n_points is not None) and (n_max is None):
+        if (
+            not isinstance(n_points, (int, np.integer)) or
+            (np.asarray(n_points).size > 1)
+        ):
+            raise ValueError(
+                "The number of points needs to be a positive natural number. ",
+                f"Instead it is {n_points}",
+            )
 
     # get number of points from required spherical harmonic order
     # ([1], chapter 3.3)
--- a/spharpy/samplings/samplings.py
+++ b/spharpy/samplings/samplings.py
@@ -491,7 +491,7 @@ def equiangular(n_points=None, n_max=None, radius=1.):
         weights = None
         quadrature = False
     else:
-        q = 2*np.arange(0, n_max + 1) + 1
+        q = 2 * np.arange(0, n_max + 1) + 1
         weights_theta = np.sin(theta_angles) * (
             1/q @ np.sin(q[np.newaxis].T @ theta_angles[np.newaxis]))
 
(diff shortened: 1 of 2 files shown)


spharpy/samplings/samplings.py
```diff
@@ -564,15 +564,14 @@ def gaussian(n_points=None, n_max=None, radius=1.):
         n_max = int(n_max)
 
     # construct the sampling grid
-    legendre, weights = np.polynomial.legendre.leggauss(int(n_theta))
+    legendre, weights = np.polynomial.legendre.leggauss(n_max+1)
```
ahms5, line 567:
```suggestion
    legendre, weights = np.polynomial.legendre.leggauss(int(n_theta))
```
i think this should be n_theta, otherwise n_theta could be wrong if points are set separatly.

spharpy/samplings/samplings.py
```diff
@@ -549,19 +555,18 @@ def gaussian(n_points=None, n_max=None, radius=1.):
         raise ValueError(
             "Either the n_points or n_max needs to be specified.")
 
+    if n_points is not None and (np.asarray(n_points).size > 1):
```
tluebeck, line 558: shouldn't we also check whether it is an integer? At the moment there would be no error for floats

tluebeck (approved):
Thanks for the fix. Looks good to me. Only one tiny comment

spharpy/samplings/samplings.py
```diff
@@ -413,7 +413,9 @@ def equiangular(n_points=None, n_max=None, radius=1.):
 
     For detailed information, see [#]_, Chapter 3.2. This is a quadrature
     sampling with the sum of the sampling weights in `sampling.weights`
-    being :math:`4\pi`.
+    being :math:`4\pi` if the number of rings is even and there number of
```
f-brinkmann, line 416: Little typo
```suggestion
    being :math:`4\pi` if the number of rings is even and the number of
```

f-brinkmann (changes requested):
Thanks, I found a little typo and would suggest to use this to also improve the docstring of equiangular n_points to something like

```
    n_points : int, tuple of two ints
        Number of sampling points in azimuth and elevation. If this
        is an integer, it sets the number of points in azimuth and
        elevation to `n_points. If this is a tuple of two integers,
        the first value sets the number of points in azimuth and the
        second in elevation.
        Either `n_points` or `n_max` must be provided. The default is
        ``None``.
```

spharpy/samplings/samplings.py
```diff
@@ -515,8 +520,9 @@ def gaussian(n_points=None, n_max=None, radius=1.):
 
     Parameters
     ----------
-    n_points : int, tuple of two ints
-        Number of sampling points in azimuth and elevation. Either `n_points`
+    n_points : int
```
xefonon, line 523: I'm just wondering if `n_points` should be named something a little more obvious, like `n_elev_points` due to what it actually represents; otherwise I see it as a bit misleading.

xefonon (commented):
Thanks, I had one comment (+1 more below) that needs your feedback. If you're ok with it, then I am too. :)

My +1 comment is: I realise we don't offer [healpix](https://healpix.sourceforge.io/) sampling of the sphere, but maybe that's for another PR.

mberz:
I very much agree that n_points is misleading and should be renamed in equiangular and Gaussian samplings.

Issue #116, opened 2025-05-06
Doc: adding SamplingSphere to documention


ahms5:
we need to add SamplingSphere to the docs, we could do it by adding a Classes section similar to pyfar.

this cameup in #115

ahms5:
edit, after 
- #99 
- #101

PR #127, opened 2025-05-28
Fix broken Sphinx links in samplings

f-brinkmann:

- Fix broken Sphinx links in Samplings

### Note

The links to SamplingSphere are broken as well, and I'm not sure why. I tried linking as we do in pyfar, but that does not work. This is marked as a ToDo in #115 and will not be done in this PR


mberz:
> The links to SamplingSphere are broken as well, and I'm not sure why. I tried linking as we do in pyfar, but that does not work as well, e.g.
> 
> ```python
> """
> Parameters
> ------------
> output : SamplingSphere
> """
> ```

This is due to this issue https://github.com/pyfar/spharpy/issues/116

f-brinkmann:
Thanks for the hint. I added this as a ToDo in #116

mberz:
Moved this to "on hold" until main is merged back into develop (after #99 #82 are merged)

ahms5 (approved):
I would say that #116 is another issue and can be adressed in a later PR.
So I would approve if merge conflicts are solved.
we could still wait for merging after #99 and #101

PR #130, opened 2025-05-30
fix history and copyrights

f-brinkmann:

Fix thinkgs that got noticed here: https://github.com/pyfar/spharpy/pull/128


ahms5 (commented):
its fine, but would be cleaner to have it in independend prs, because both changes have nothing todo with eachother

Issue #122, opened 2025-05-20
Spharpy Roadmaps v1.0.0


f-brinkmann:
# 1.0.0 (merges into dev)

- [ ] Code style - wait until other stuff is merged (#147)
- [ ] Deprecations
  - [ ] #143 (and wait until scipy bug is fixed?)
  - [ ]  #142 
- [ ] SH classes
  - [x] SH Class #101 
  - [x] SH Signal Class #99
  - [x] SH conversion methods #113
- [ ] SamplingSphere
  - [x] Enforce pyfar>=0.7.3 #124 
  - [x] Add check for unique radius #121
  - [x] Add quadrature property #120
  - [x] Clean tests (there are often two tests for one sampling because old tests from pyfar have been copied but not merged) #134 
- [ ] Samplings
  - [ ] rename `n_points` to something more meaningful #131 
  - [ ] #135
- [x] merge main into dev and resolve conflicts #128
- [ ] #153 
- [ ] Documentation (after merging #128)
  - [ ] Check and add documentation for sampling sphere (see #137, #116)
  - [ ] For the SH classes, mention which parameters 'basis_type', 'normalization', 'condon_shortley' must be set to follow common conventions (Ambix, Williams/Rafaely, Ahrens, more?)
  - [ ] In the readme, replace the broken link [example notebook](https://github.com/pyfar/pyfar/blob/main/examples) with a link to the examples section of the gallery

f-brinkmann:
> I would remove the Deplyong and Testing section here, because this should be done as a hotfix on main first, because it is also relavent for main.
> 
> just because we are developing v1.0.0 now, we still need to maintain the current versions

Good point, made it two separate lists

mberz:
I agree that the milestones in project boards are better suited than issues. I've copied the content to the milestone, but it needs cleaning and updates anyway. Closing the issue.

PR #134, opened 2025-05-30
Clean samplings tests

f-brinkmann:

Part of #122 

The tests were partly redundant because they were copied from pyfar. This redundancy was removed and mentions of pyfar were removed.

### DIFF
--- a/tests/test_samplings.py
+++ b/tests/test_samplings.py
@@ -19,7 +19,7 @@ def test_cube_equidistant_int():
     np.testing.assert_allclose(x, coords.x)
     np.testing.assert_allclose(y, coords.y)
     np.testing.assert_allclose(z, coords.z)
-    assert isinstance(coords, Coordinates)
+    assert type(coords) is Coordinates
     assert coords.csize == 3**3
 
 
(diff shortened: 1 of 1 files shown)


tests/test_samplings.py
```diff
@@ -19,14 +19,11 @@ def test_cube_equidistant():
     np.testing.assert_allclose(x, coords.x)
     np.testing.assert_allclose(y, coords.y)
     np.testing.assert_allclose(z, coords.z)
+    assert isinstance(coords, Coordinates)
```
ahms5, line 22:
```suggestion
    assert type(coords) is Coordinates
```
shoudnt we be more explict here? due to the SamplingSphere and Coordinates inheritance?

tests/test_samplings.py
```diff
@@ -322,7 +313,7 @@ def test_gaussian_pyfar():
 
     # test with single number of points
     c = samplings.gaussian(5)
-    isinstance(c, Coordinates)
+    isinstance(c, SamplingSphere)
```
ahms5, line 316:
```suggestion
    assert isinstance(c, SamplingSphere)
```
otherwise nothing happens, hope it is not also True for Coordinates

ahms5 (changes requested):

tests/test_samplings.py
f-brinkmann: Good point. I've replaced all `isinstances` with `type` to be future-proof

tests/test_samplings.py
f-brinkmann: Thanks for catching. I found a few more occurances of this and corrected them all.

ahms5 (approved):

PR #135, opened 2025-05-30
Rename spherical_t_design

f-brinkmann:

Rename `spherical_t_design` from `spharpy.samplings` to `t_design`. I think the old name was a left-over from moving samplings from pyfar to spharpy. All other samplings had the `spherical` already removed. Or is there a non-spherical t-design, as well?


ahms5 (changes requested):
thank you for taking care, we could also update the docstring: rename n_sh into n_max.

xefonon (approved):
Thanks, looks good to me.

PR #136, opened 2025-06-01
Bug: Doc: clean up imports in doc

ahms5:

- removed ``indexing`` module from doc 
- shifted special reference to modules folder and included it into the doctree
- ci: doc test is now failing if warning happens
- requires #138 for ci to pass

edit 1.6.25:
- added ``indexing`` module to docs accroing to @mberz review
- fixes footnotes warnings to prevent test to fail
  - changes were made, sothat the first sentence of the docstring does not contain a reference, this leads to problems within autofootnote.

edit 25.7.
- What would be the best place in the documentation for `indexing` and `spherical`

edit 12.9.
- all submodules are also in the doc

edit 19.9.:
- ruff is failing due to #189, it is not related to this pr

### DIFF
--- a/spharpy/spherical.py
+++ b/spharpy/spherical.py
@@ -463,7 +463,8 @@ def spherical_harmonic_basis_gradient(n_max, coordinates, normalization="n3d",
     r"""
     Calculates the unit sphere gradients of the complex spherical harmonics.
 
-    See [#]_ for details.
+    This implementation avoids singularities at the poles using identities
+    derived in [#]_.
 
     The angular parts of the gradient are defined as
 
@@ -476,13 +477,8 @@ def spherical_harmonic_basis_gradient(n_max, coordinates, normalization="n3d",
         {\partial \theta} \vec{e}_\theta .
 
 
-    This implementation avoids singularities at the poles using identities
-    derived in [#]_.
-
-
     References
     ----------
-    .. [#]  E. G. Williams, Fourier Acoustics. Academic Press, 1999.
     .. [#]  J. Du, C. Chen, V. Lesur, and L. Wang, “Non-singular spherical
             harmonic expressions of geomagnetic vector and gradient tensor
             fields in the local north-oriented reference frame,” Geoscientific
@@ -663,8 +659,8 @@ def spherical_harmonic_basis_gradient_real(n_max, coordinates,
     Calculates the unit sphere gradients of the real valued spherical
     harmonics.
 
-    The spherical harmonic functions are fully normalized (N3D) and follow
-    the AmbiX phase convention [#]_.
+    This implementation avoids singularities at the poles using identities
+    derived in [#]_.
 
     The angular parts of the gradient are defined as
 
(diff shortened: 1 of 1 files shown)


f-brinkmann (commented):
We should maybe discuss if `spharpy.sphecial` should be considered a private module. For example the spherical_harmonics basis methods are barebone wraps of corresponding scipy methods with less documentation than in the SH classes. As far as I understand it, they are used as helpers only and such similar functions defined in multiple places might be confusing.

docs/api_reference.rst
```diff
@@ -17,12 +17,11 @@ Modules
    :maxdepth: 1
 
    modules/spharpy.beamforming
-   modules/spharpy.indexing
    modules/spharpy.interpolate
    modules/spharpy.plot
    modules/spharpy.samplings
    modules/spharpy.spatial
-   modules/spharpy.spherical
+   modules/spharpy.special
```
f-brinkmann, line 24: is it intended that spherical was removed and special stays?

ahms5: upps

mberz (changes requested):
I disagree here. If we remove it from the docs people typically do not know about the existence of it. So please re-add it.

ahms5:
I agree, we should add what we have. should we simply close it and add the ci doc tests with warning i a dedicated pr ?

ahms5:
- added ``indexing`` module to docs accroing to @mberz review
- fixes footnotes warnings to prevent test to fail
  - changes were made, sothat the first sentence of the docstring does not contain a reference, this leads to problems within autofootnote.
  
  ready for review

f-brinkmann (approved):

docs/api_reference.rst
```diff
@@ -23,13 +23,13 @@ Modules
 
 .. toctree::
    :maxdepth: 1
-
+
```
mberz, line 26: Please do not add blank spaces (if you use the editorconfig plugin it will remove trailing whitespaces automatically ;))

spharpy/spherical.py
```diff
@@ -472,7 +476,7 @@ def spherical_harmonic_basis_gradient(n_max, coordinates, normalization="n3d",
 
 
     This implementation avoids singularities at the poles using identities
-    derived in [#]_.
+    derived in [#]_ and [#]_.
```
mberz, line 479: This is incorrect, the singularity avoiding formulation is detailed only in the second reference.

mberz (changes requested):

spharpy/spherical.py
ahms5: @mberz I tried to fix it. But I'm not into the topic, if this isnt correct, or you have an better idea, feel free to change here in the PR directly

docs/modules/spharpy.special.rst
```diff
@@ -1,7 +1,8 @@
 spharpy.special
------------------
+---------------
 
 .. automodule:: spharpy.special
    :members:
+   :special-members: __init__
```
hoyer-a, line 6: I'm wondering why this was added, since currently there are no classes inside special

hoyer-a (approved):
Just one question, otherwise approved.

mberz (approved):

PR #143, opened 2025-06-01
Deprecate scipy Legendre function

f-brinkmann:

**Note the following**: Failling tests are not caused by this PR

- Use scipy's new `accoc_legendre` instead of `lpmn` which will be deprecated in scipy 1.17
- Add tests for spharpy.special.legendre_function to make sure changes work as intended
- Restriction scipy to version >= 1.17.0 is required because of https://github.com/scipy/scipy/issues/23101
- Restriction to Python >= 3.11 required because scipy 1.17 is not available for this version
- Minor corrections in the docstrings

### DIFF
--- a/tests/test_special.py
+++ b/tests/test_special.py
@@ -278,9 +278,10 @@ def test_spherical_harmonic_gradient_phi_real():
         npt.assert_allclose(actual, desired, rtol=1e-10, atol=1e-10)
 
 
-@pytest.mark.parametrize(['condon_shortley'], [(True, ), (False, )])
-@pytest.mark.parametrize(['m'], [(-1, ), (0, ), (1, )])
-def test_legendre(condon_shortley, m):
+@pytest.mark.parametrize('condon_shortley', [(True, ), (False, )])
+@pytest.mark.parametrize(('n', 'm'), [
+    (1, -1), (1, 0), (1, 1), (2, -2), (2, -1), (2, 0), (2, 1), (2, 2)])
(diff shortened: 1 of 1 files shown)


f-brinkmann:
> If it requires a certain scipy version, we should add it to the pyproject.toml

It scheduled for scipy 1.17, but there hasn't been much action on the issue. I added a todo to the top so we don't forget.

mberz:
There's an update on the respective scipy issue, let's wait for further updates here https://github.com/scipy/scipy/issues/23101

mberz:
I've created a PR to fix the underlying issue with the Legendre functions in scipy.
In any case, the issue is scheduled to be fixed no later than scipy 1.17.0, which is when `scipy.special.lpmv` is to be removed.
So it's probably a good idea to wait with this PR.

mberz (approved):

tests/test_special.py
```diff
@@ -276,3 +276,39 @@ def test_spherical_harmonic_gradient_phi_real():
...
+    if m == 1:
+        desired =  -np.sqrt(1 - z**2)
+
+    # remove Condon-Shortley phase
+    if not condon_shortley and m % 2:
+        desired *= -1
+
+    # compute and compare actual values
+    actual = special.legendre_function(1, m, z, condon_shortley)
+    npt.assert_almost_equal(actual, desired, 10)
```
mberz: Just realized that this test does nothing, since m is 1 at max, and no even m is used in the test.

PR #144, opened 2025-06-02
Fix secondary sidebar display of class attributes

hoyer-a:
Adjust display of secondary sidebar using custom js as in https://github.com/pyfar/pyfar/pull/788
Show only class `attribute` in sidebar instead of `class.attribute`.


ahms5 (commented):
Thank you! would it make sence to generally pull the custum.js from the gallery via the doc build?

ahms5:
Yes, maybe we can create a pr in gallery to add the file.  We should add it in the cookiecutter then and then we apply it here and in any other package. 
This might take some time. So I would be fine to merge it here right now, and to replace it later on with the cookiecutter version.

ahms5 (approved):

ahms5 (approved):

PR #145, opened 2025-06-02
restrict scipy version, due to deprecations

### LINKED ISSUE #83: `scipy.special.sph_harm` is deprecated
[sph_harm](https://docs.scipy.org/doc/scipy-1.15.0/reference/generated/scipy.special.sph_harm.html) will be removed in scipy 1.17. 
The new sh-function is [sph_harm_y](https://docs.scipy.org/doc/scipy-1.15.0/reference/generated/scipy.special.sph_harm_y.html#scipy.special.sph_harm_y)

ahms5:

Closes #83

- this a fix for the main branch.
- we dont want to force scipy 1.15, so we restict the scipy version before the deprecation happens
-

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ classifiers = [
 ]
 dependencies = [
     'numpy>=1.22',
-    'scipy',
+    'scipy<1.18.0',
     'urllib3',
     'matplotlib>=3.10.3',
     'pyfar<0.8.0',


pyproject.toml
```diff
@@ -26,7 +26,7 @@ classifiers = [
 ]
 dependencies = [
     'numpy>=1.22',
-    'scipy',
+    'scipy<1.18.0',
```
f-brinkmann, line 29: If we do this, we will always get the deprecation warnings, alternatively we could do
```suggestion
    'scipy<1.15.0',
```
to avoid this. Note that there might currently be a scipy bug, that would not be fixed in 0.6.3: https://github.com/scipy/scipy/issues/23101

f-brinkmann (commented):
Left a note for discussion

mberz:
The functions will be removed in 1.17, so allowing 1.17 does not solve the issue as far as I see it.

Closing in favor of #180 (I saw this PR after the fact, sorry about that)

PR #152, opened 2025-06-03
Spherical harmonic transform

### LINKED ISSUE #133: Add Spherical Harmonic Transforms
After introducing the new SH and SHSignal classes, we should think of how we want to implement SH (inverse) transforms

tluebeck:

Closes #133

- add spherical harmonics transform and its inverse
- depends on #101 
- we need to decide where we implement it, probably shift to `transforms`?


spharpy/sht.py
```diff
@@ -0,0 +1,124 @@
+import numpy as np
+from pyfar import Signal
+import warnings
+from . import SphericalHarmonicSignal
+from . import SphericalHarmonics
+
+
+def sht(signal, coordinates, n_max, basis_type="real", axis=-2,
+        channel_convention='acn', normalization='n3d', condon_shortley='auto',
+        inverse_method='pseudo_inverse'):
```
mberz, line 10:
```suggestion
def sht(signal, spherical_harmonics):
```
How about passing a SphericalHarmonic object here. This simplifies things a lot.

mberz (changes requested):
I think we should review the concept again. Specifying all properties again does not use the power of the SH class at all from a users perspective.
Passing the SH class instead of creating it within the function would make things much easier in my opinion.

spharpy/sht.py
tluebeck: I think it is easier for the user to only have one method which creates the SH class and performs the transform. Moreover, with the default parameters the user only needs to pass the signal, coordinates and n_max. But I thinks its better to discuss in the weekly.

mberz: Yes, let's discuss during the weekly.

f-brinkmann:
- We decided that the functions should get an SH class object.
- @tluebeck will check if it is easy to have SHTimeData and SHFrequencyData by changing inheritance.

spharpy/sht.py
```diff
@@ -0,0 +1,138 @@
...
+    """
+
+    if not signal.time.shape[axis] == coordinates.csize:
+        if coordinates.csize not in signal.time.shape:
+            raise ValueError("Signal shape does not match "
+                             "number of coordinates.")
+        else:
+            axis = signal.time.shape.index(coordinates.csize)
+            warnings.warn("Compute spherical harmonics transform along "
+                          f"axis = {axis}.", UserWarning)
```
f-brinkmann, line 65: I think we should only raise an error here and not change the user input. But I like the idea of detecting the axis. How about having `axis='auto'` as a default and use

`axis = np.where(np.array(signal.cshape) == Y_inv.shape[1])[0]` 

to detect the axis. If `len(axis) == 1` we proceed if `len(axis) == 0` we raise an error and if `len(axis) > 1` we raise a different errror.

f-brinkmann (commented):
Thanks for the effort. Functionality is great to have and looks good to me. I only marked two points for reducing code for now.

f-brinkmann (commented):
Code and tests look good to me.

The new functions don't show up in the docs currently and I think it might be good to provide a mathematical definition of the SHT and ISHT somewhere. I see three options and list them in order of my preference. In any case, this could be done after merging the current state for east of review.

1. Add a spharpy.sht module, that contains sht and isht and a top-level module docstring.
2. Add a section 'Spherical harmonic transform' to the 'Definitions' sections of the spharpy docs and reference that section in sht and isht
3. Extend the documentation of sht and isht.

mberz:
To me adding it to the module docstring is probably the most intuitive version. 
I would include it in this PR, rather than a separate one.

tluebeck:
I added the documentation to the top of the module and added it to the docs.

mberz:
waiting for the outcome of the discussion in #326

PR #156, opened 2025-06-06
Add check if sampling is quadrature

### LINKED ISSUE #155: Stricter check of quadrature property in SamplingSphere
@xefonon, @ahms5, @mberz @tluebeck , I'm sorry but I can't let go of the handling for `quadrature` and think we should reconsider our decision. Because we have a basis matrix Y and weights w, we have a clear mathematical criterion to check if we have a quadrature sampling, which is to test if `Y.T @ diag(w) @ Y` equals the identity matrix. This would mean that `SpharicalHarmonics.quadrature` could be a getter only property that checks

```python
import spharpy as sy
import numpy as np
import numpy.testing as npt

# generate a sampling
quadrature = True
if quadrature:
    sampling = sy.samplings.gaussian(n_max=3)
else:
    sampling = sy.samplings.hyperinterpolation(n_max=3)

# generate SH object
sh = sy.spherical.SphericalHarmonics(n_max=3, coordinates=sampling)

# test if basis is quadrature
n_coeffs = (sh.n_max + 1)**2
basis = sh.basis[:, :n_coeffs]
quadrature_test = basis.T @ np.diag(sh.coordinates.weights) @ basis
diag = np.diag(np.ones(n_coeffs))

# can not use relative tolerance when testing for identity matrix due to zeros
decimal = 10  # precision could be discussed or a class paramneter
npt.assert_almost_equal(quadrature_test, diag, decimal)
```

This would make things easier and more clear I think

- `quadrature` could be removed from init
- if setting `inverse_transform_method = 'quadrature'` the setter checks `SphericalHarmonics.quadrature` and raises an error if the criterion is not met.

_Originally posted by @f-brinkmann in https://github.com/pyfar/spharpy/pull/101#pullrequestreview-2887125229_

tluebeck:

Closes #155 

- add a check if sampling is a valid quadrature 
- add property `quadrature_tolerance`
- add private method `_check_quadrature`
- update quadrature setter
- do we want to do a quadrature check in the weights_setter?


spharpy/classes/coordinates.py
```diff
@@ -75,6 +85,12 @@ def __init__(
             radius and an error is raised if the maximum deviation from the
             mean radius exceeds this tolerance. The default of ``1e-6`` meter
             is intended to allow for some numerical inaccuracy.
+        quadrature_tolerance : float, optional
+            Tolerance for testing whether the provided sampling grid is
+            a valid quadrature. The sampling is considered a valid quadrature
+            if the deviation of the inner product of the Spherical Harmonics
```
f-brinkmann, line 91:
```suggestion
            if the deviation of the inner product of the weighted Spherical Harmonics
```

spharpy/classes/coordinates.py
```diff
@@ -10,17 +10,24 @@
 :py:attr:`~spharpy.SamplingSphere.weights` are not None, their sum must
 equal the integral over the unit sphere, which is :math:`4\pi`.
 
-It also adds two additional properties:
+The flag :py:attr:`~spharpy.SamplingSphere.quadrature` specifies if the
+points belong to a quadrature, which requires that the
+:py:attr:`~spharpy.SamplingSphere.weights` sum to :math:`4 \pi`, the maximum
+spherical harmonic order of the sampling
+grid :py:attr:`~spharpy.SamplingSphere.n_max` is specified and the inner
+product of the Spherical Harmonics matrix :math:`\mathrm{Y}`
```
f-brinkmann, line 18:
```suggestion
product of the weighted Spherical Harmonics matrix :math:`\mathrm{Y}`
```

spharpy/classes/coordinates.py
```diff
@@ -128,6 +147,12 @@ def from_cartesian(
             radius and an error is raised if the maximum deviation from the
             mean radius exceeds this tolerance. The default of ``1e-6`` meter
             is intended to allow for some numerical inaccuracy.
+        quadrature_tolerance : float, optional
+            Tolerance for testing whether the provided sampling grid is
+            a valid quadrature. The sampling is considered a valid quadrature
+            if the deviation of the inner product of the Spherical Harmonics
```
f-brinkmann, line 153:
```suggestion
            if the deviation of the inner product of the weighted Spherical Harmonics
```

spharpy/classes/coordinates.py
```diff
@@ -183,6 +210,12 @@ def from_spherical_elevation(
...
```
f-brinkmann, line 216:
```suggestion
            if the deviation of the inner product of the weihghted Spherical Harmonics
```

spharpy/classes/coordinates.py
```diff
@@ -238,6 +273,12 @@ def from_spherical_colatitude(
...
```
f-brinkmann, line 279:
```suggestion
            if the deviation of the inner product of the weighted Spherical Harmonics
```

spharpy/classes/coordinates.py
```diff
@@ -291,6 +334,12 @@ def from_spherical_side(
...
```
f-brinkmann, line 340:
```suggestion
            if the deviation of the inner product of the weighted Spherical Harmonics
```

spharpy/classes/coordinates.py
```diff
@@ -345,6 +396,12 @@ def from_spherical_front(
...
```
f-brinkmann, line 402:
```suggestion
            if the deviation of the inner product of the weighted Spherical Harmonics
```

spharpy/classes/coordinates.py
```diff
@@ -398,6 +457,12 @@ def from_cylindrical(
...
```
f-brinkmann, line 463:
```suggestion
            if the deviation of the inner product of the weighted Spherical Harmonics
```

spharpy/classes/coordinates.py
```diff
@@ -501,6 +590,27 @@ def _check_weights(self, weights):
...
 
+    def _check_quadrature(self):
+        r"""Check if the sampling is a are valid quadrature.
+
+        Returns
+        -------
+        check : bool
+            Flag which indicates if quadrature is a valid quadrature
+        """
+        if self.n_max is None or self.weights is None:
```
f-brinkmann, line 601: Not very likely to happen, but we can't check this way for zeroth order grids:
```suggestion
        if self.n_max in [None, 0] or self.weights is None:
```

spharpy/classes/coordinates.py
```diff
@@ -501,6 +590,27 @@ def _check_weights(self, weights):
...
+            return False
+        # generate SH object at N=1
+        n_eval = 1
+        sh_basis = spherical_harmonic_basis_real(n_eval, self)
+
+        # test if basis is quadrature
+        quad_evaluation = sh_basis.T @ np.diag(self.weights) @ sh_basis
+        identity = np.eye((n_eval + 1)**2)
+
+        error = np.abs(np.sum(np.sum(quad_evaluation-identity)))
```
f-brinkmann, line 611: I would rather check for the max abs deviation. Currently positive and negative deviations could cancel out when taking the sum before the abs
```suggestion
        error = np.max(np.abs(quad_evaluation-identity))
```

f-brinkmann (changes requested):
Thanks, I mostly have very small doc string or structural suggestions. Looks almost good to me. And the tests directly told us that classified the Fliege grid as quadrature by mistake. I'll open a separate pull for this.

spharpy/classes/coordinates.py
tluebeck: Are we sure a 0-order grid isn't a quadrature, at least in the mathematical sense? If so, we need need to change something in the gaussian grid ( if n_max == 0, quadrature = false or so).

f-brinkmann: True, it would be in the mathematical sense. Probably useless in practice - but I agree to leave everything as is to be clean.

spharpy/classes/coordinates.py
```diff
@@ -501,6 +590,27 @@ def _check_weights(self, weights):
 
         return weights
 
+    def _check_quadrature(self):
+        r"""Check if the sampling is a are valid quadrature.
```
ahms5, line 594:
```suggestion
        r"""Check if the sampling is a valid quadrature.
```

f-brinkmann:
@tluebeck @mberz @ahms5 I think the changes are good but openend a more general thought to me: Any changes in the SamplingSphere after setting and checking `quadrature` could change the sampling from being a quadtrature to not being a quadrature any more. Since we are now using an analytical expression to check for this, do we need a setter anymore? If not things would be simpler. We could make `quadrature`a getter-only property and check each time `SamplingSphere.quadrature` is called.

spharpy/classes/coordinates.py
```diff
@@ -291,6 +334,12 @@ def from_spherical_side(
             radius and an error is raised if the maximum deviation from the
             mean radius exceeds this tolerance. The default of ``1e-6`` meter
             is intended to allow for some numerical inaccuracy.
+        quadrature_tolerance : float, optional
+            Tolerance for testing whether the provided sampling grid is
+            a valid quadrature. The sampling is considered a valid quadrature
+            if the deviation of the inner product of the weighted Spherical
+            Harmonics matrix from the identity matrix is smaller than the
```
mberz, line 341:
```suggestion
            if the deviation of the inner product of the weighted spherical
            harmonics matrix from the identity matrix is smaller than the
```

spharpy/classes/coordinates.py
```diff
@@ -128,6 +147,12 @@ def from_cartesian(
...
```
mberz, line 154:
```suggestion
            if the deviation of the inner product of the weighted spherical
            harmonics matrix from the identity matrix is smaller than the
```

spharpy/classes/coordinates.py
```diff
@@ -75,6 +85,12 @@ def __init__(
...
```
mberz, line 92:
```suggestion
            if the deviation of the inner product of the weighted spherical
            harmonics matrix from the identity matrix is smaller than the
```

f-brinkmann:
We decided that
- `quadrature` should be a getter only
- it will be computed if requested and it will be cached (i.e recomputed only if required)
- the property will test for `n_max`
- in testing, we only check it for 1st order

spharpy/classes/coordinates.py
```diff
@@ -511,28 +594,20 @@ def weights(self):
     @weights.setter
     def weights(self, weights):
         r"""The area/quadrature weights of the sampling.
-        Their sum must equal to :math:`4\pi`.
+        Their sum must equal to :math:`4\pi`. If the provided
+        weights do not fulfill the quadrature condition,
+        the quadrature flag will be updated accordingly.
```
f-brinkmann, line 599: Should be identical for getter and setter. I would omitt the information about the quadrature flag.

```suggestion
        r"""Get or set the area/quadrature weights of the sampling.
        Their sum must equal to :math:`4\pi`.
```

f-brinkmann (approved):
Thanks for the update. I only had a minor suggestion for a docstring. Otherwise approved.

tests/test_coordinates.py
```diff
@@ -151,29 +156,24 @@ def test_setting_weights_invalid():
...
+
+def test_quadrature_getter_changing_points():
+    # create a quadrature grid (gaussian)
+    sampling = gaussian(n_max=3)
+    # check if quadrature is set properly
+    assert sampling.quadrature
+    # update points such that quadrature is not valid anymore
+    sampling.set_sph(np.random.rand(sampling.csize, 1),
+                     np.ones((sampling.csize, 1)),
+                     np.ones((sampling.csize, 1)))
```
ahms5, line 178:
```suggestion
    sampling.spherical_colatitude = np.array(np.random.rand(sampling.csize, 1),
                     np.ones((sampling.csize, 1)),
                     np.ones((sampling.csize, 1)))
```
sth like this should work
I think this is deprecated and will be removed in 0.8.0.

ahms5 (commented):
just one deprecation in the tests, otherwise approved

tests/test_coordinates.py
ahms5: and fix ruff

ahms5 (commented):
seems like ruff is failing for some reason

mberz (changes requested):
Thanks! Mostly docstring comments.
The comment on the property getter is more of a question.

PR #157, opened 2025-06-06
cast n_max to int in Hyperinterpolation

tluebeck:

- cast n_max to int in `hyperinterpolation` in case n_points are passed


mberz (approved):
Looks good, thanks!

PR #151, opened 2025-06-03
[On Hold] add SHSignal to docs

tluebeck:
### Which issue(s) are closed by this pull request?

Closes #

### Changes proposed in this pull request:

- added SphericalHarmonicSignal to documentation, similar to pyfar style
- added introduction to SpamplingSphere documentation page
- we need to wait for #101 to add a general documentation of the Spherical Harmonics implementation

PR #160, opened 2025-06-15
Move SH class to sphary.classes

f-brinkmann:

required for #151 

- Move SH class to sphapy.classes
- Change imports and related code
- I commited with `--author=xefonon` so I hope the original authorship is maintained


xefonon (approved):
Thanks, it looks fine to me. :)

f-brinkmann:
ok - merging this and @tluebeck can maybe move the SH Signal class to the same file in a separate pull. After this we can continue updating the docs, where Tim already opened a pull :)

PR #166, opened 2025-07-06
Add SphericalHarmonicsAudio classes

### LINKED ISSUE #165: SH Time and FrequencyData
Let's explore how to easily adopt the added properties for SphericalHarmonicSignal to TimeData and FrequencyData.

tluebeck:

Closes #165 

- introduce SphericalHarmonicsAudio base class which holds all SH related properties
- introduce SphericalHarmonicsTimeData and FrequencyData


f-brinkmann (commented):
The import of SH TimeData and FrequencyData is missing in `spharpy/__init__.py`

spharpy/classes/audio.py
```diff
@@ -1,12 +1,208 @@
...
+    data : array, double
+        Raw data in the frequency domain. The data should have at least
+        3 dimensions, according to the 'C' memory layout, e.g. data of
+        ``shape = (1, 4, 1024)`` has 1 channel with 4 spherical harmonic
+        coefficients with 1024 frequency bins each.. Data can be ``int``,
+        ``float`` or ``complex``. Data of type ``int`` is converted to
+        ``float``.
+    frequencies : array, double
+        Frequencies of the data in Hz. The number of frequencies must match
+        the size of the last dimension of data.
```
f-brinkmann:
```suggestion
        the size of the last dimension of `data`, i.e., ``data.shape[-1]``.
```

spharpy/classes/audio.py
```diff
@@ -1,12 +1,208 @@
...
+    basis_type : str
+        Type of spherical harmonic basis, either ``'complex'`` or
+        ``'real'``.
+    normalization : str
+        Normalization convention, either ``'N3D'``, ``'NM'``,
+        ``'maxN'``, ``'SN3D'`` or ``'SNM'``. (maxN is only supported up
+        to 3rd order)
+    channel_convention : str
+        Channel ordering convention, either ``'acn'`` or ``'fuma'``.
```
f-brinkmann:
```suggestion
        Channel ordering convention, either ``'ACN'`` or ``'FuMa'``.
```

spharpy/classes/audio.py
```diff
@@ -1,12 +1,208 @@
...
+        (FuMa is only supported up to 3rd order)
+    condon_shortley : bool
+        Flag to indicate if the Condon-Shortley phase term is included
+        (``True``) or not (``False``).
+    comment : str
+        A comment related to `data`. The default is ``None``.
```
f-brinkmann:
```suggestion
        A comment related to `data`. The default is ``""``.
```

mberz:
> Sorry, forgot one thing. I would suggest to rename `audio.py` to `sh_audio.py`.

Please rename files in this pr. this will cause conflicts with other branches.

mberz:
> > Sorry, forgot one thing. I would suggest to rename `audio.py` to `sh_audio.py`.
> 
> Please rename files in this pr. this will cause conflicts with other branches.

I'm very sorry, I meant to write don't rename in this branch, since the file is already part of develop. 
Could you please revert dfd4d547220efbd4eb03a37fcea2210c7f39eb21 ?

mberz (commented):
I'm just commenting on this now, as I feel we need to make some adjustments to the SphericalHarmonicDefinition class before continuing with this PR.

1. I think it makes sense to move the functionality of saving the old properties to the parent class, as it could come in handy in other classes. 
2. I now remember why I first had the SphericalHarmonicDefintion in mind without the `n_max` property defined. I'll add an abstract base class for that which could serve as parent for the SphericalHarmonicAudio classes.

Sorry to postpone this PR once again.

spharpy/classes/audio.py
```diff
@@ -1,34 +1,27 @@
...
-    """Create audio object with spherical harmonics coefficients in time or
-    frequency domain.
+class _SphericalHarmonicAudio(_Audio):
+    """Base class for spherical harmonics audio objects.
+
+    This class extends the pyfar Audio class with all methods and
+    properties required for spherical harmonics data and are common to the 
+    three sub-classes :py:func:`SphericalHarmonicsTimeData`,
+    :py:func:`SphericalHarmonicsFrequencyData`, and
+    :py:func:`SphericalHarmonicsSignal`.
```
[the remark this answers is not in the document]
tluebeck: Sure, reverted. I kind of had the feeling it was a typo cause it did't made sense to me :D ... 
I agree that it makes sense to move the saving of the old parameters to the SHDefinitions class. However, I think it could be an option to do that in a separate PR. The functionality from outside would not change, and we could move on with other stuff, like SHtrafo etc.

spharpy/classes/audio.py
```diff
@@ -178,3 +141,182 @@ def channel_convention(self, value):
...
+    channel_convention : str
+        Channel ordering convention, either ``'acn'`` or ``'fuma'``.
+        (FuMa is only supported up to 3rd order)
+    condon_shortley : bool
+        Flag to indicate if the Condon-Shortley phase term is included
+        (``True``) or not (``False``).
+    comment : str
+        A comment related to `data`. The default is ``None``.
+    """
+    def __init__(self, data, frequencies, basis_type, normalization,
```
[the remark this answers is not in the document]
tluebeck: Yes, I would suggest to not use defaults here.

f-brinkmann (commented):
Following up on the question on slack, I think it would be cleaner to keep `data` out of `_SHAudio`. Could this work:

- move checks for `n_max` from `_SHAudio.__init__` to a private method, e.g.,  `_SHAudio._check_data_shape`. This avoids the need to pass `data` to init.
- Call this private method in each setter, e.g., `SHTimeData.time`
- In each init, e.g., `SHTimeData` first initialize `_SHAudio`, than call the private method for checking the data shape, and finally call `TimeData.__init__`

Other points:

Instead of inheriting from `_SphericalHarmonicBase` and re-implementing the `n_max` getter, I think it would be cleaner to inherit from `SphericalHarmonicDefinition` and
- overload the `n_max` setter to raise a ValueError, e.g., 'n_max can not be set and is defined by the shape of the data contained in the spherical harmonic Audio object' 
- overload the `basis_type` setter to raise a ValueError, e.g., 'Setting the basis_type is not yet supported'. This could be implemented at a later point, if we want. Since we decided to internally store data in standard conventions, this would be fairly straigt forward in this case.

spharpy/classes/audio.py
```diff
@@ -1,12 +1,280 @@
...
+        Domain of data. The default is ``'time'``
+    comment : str
+        A comment related to `data`. The default is ``None``.
+
+    """
+    def __init__(self, basis_type, normalization, channel_convention,
+                 condon_shortley):
+
+        # check dimensions
+        if len(self._data.shape) < 3:
```
mberz:
```suggestion
        if len(self.cshape) < 2:
```

spharpy/classes/audio.py
```diff
@@ -1,12 +1,280 @@
...
+            value = change_channel_convention(
+                        value,
+                        self.channel_convention,
+                        "ACN",
+                        axis=-2)
+
+        FrequencyData.freq.fset(self, value)
+
+
+class SphericalHarmonicSignal(_SphericalHarmonicAudio, Signal):
```
sikersten: Maybe I do not fully get the concept, but shouldn't this be
```suggestion
class SphericalHarmonicSignal(SphericalHarmonicTimeData, SphericalHarmonicFrequencyData):
```

spharpy/classes/audio.py
```diff
@@ -1,12 +1,276 @@
...
+
+class _SphericalHarmonicAudio(_Audio, _SphericalHarmonicBase, ABC):
+    """
+    Base class for spherical harmonics audio objects.
+
+    This class extends the pyfar Audio class with all methods and
+    properties required for spherical harmonics data and are common to the
+    three sub-classes :py:func:`SphericalHarmonicsTimeData`,
+    :py:func:`SphericalHarmonicsFrequencyData`, and
+    :py:func:`SphericalHarmonicsSignal`.
```
ahms5:
```suggestion
    three sub-classes :py:class:`SphericalHarmonicsTimeData`,
    :py:class:`SphericalHarmonicsFrequencyData`, and
    :py:class:`SphericalHarmonicsSignal`.
```

spharpy/classes/audio.py
```diff
@@ -1,12 +1,314 @@
...
+            self,
+            basis_type,
+            channel_convention,
+            normalization,
+            condon_shortley)
+
+    @property
+    def n_max(self):
+        """Get or set the spherical harmonic order."""
+        return int(np.sqrt(self.cshape[-1])-1)
```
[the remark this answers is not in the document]
tluebeck: I'll take a closer look at that point. I tried this real quick, but if we overwrite it, we cannot init _SphericalHarmonicBase because it calls the basis_type setter :D Let me check if there is a way without changing _SphericalHarmonicBase.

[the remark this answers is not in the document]
tluebeck: Found a solution - not super beautiful but works

spharpy/classes/audio.py
```diff
@@ -1,12 +1,328 @@
...
+
+    @property
+    def basis_type(self):
+        """Get or set the type of spherical harmonic basis."""
+        return _SphericalHarmonicBase.basis_type.fget(self)
+
+    @basis_type.setter
+    def basis_type(self, value):
+        """Get or set the type of spherical harmonic basis."""
```
f-brinkmann: Its possible to overwrite only the setter, which reduced duplicate Code:
```suggestion
    @_SphericalHarmonicBase.basis_type.setter
    def basis_type(self, value):
        """Get or set the type of spherical harmonic basis."""

        # Make sure that the basis type can only be set during initialization.
        # Changing it afterwards requires implementing the conversion between
        # real and complex-valued coefficients, which is possible but not yet
        # implemented.
```

mberz (changes requested):
Thanks for the implementation.
I don't have a lot of time at the moment, so I just briefly went over the code and pointed out a couple of things I spotted right away.

spharpy/classes/audio.py
```diff
@@ -1,12 +1,328 @@
...
+    @property
+    def n_max(self):
+        """Get or set the spherical harmonic order."""
+        return int(np.sqrt(self.cshape[-1])-1)
+
+    @property
+    def basis_type(self):
+        """Get or set the type of spherical harmonic basis."""
+        return _SphericalHarmonicBase.basis_type.fget(self)
+
```
ahms5:
```suggestion
```

I think since this is doing exactly the same as the baseclass, we can simply remove it.

spharpy/classes/audio.py
```diff
@@ -1,12 +1,328 @@
...
+        (``True``) or not (``False``).
+    comment : str
+        A comment related to `data`. The default is ``""``.
+    is_complex : bool, optional
+        A flag which indicates if the time data are real or complex-valued.
+        The default is ``False``.
+    """
+
+    def __init__(self, data, times, basis_type, normalization,
+                 channel_convention, condon_shortley, comment="",
```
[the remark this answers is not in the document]
tluebeck: I agree, I'd force the user to specify the SH properties

spharpy/classes/audio.py
```diff
@@ -183,10 +189,13 @@ class SphericalHarmonicTimeData(_SphericalHarmonicAudio, TimeData):
...
-        Raw data in the time domain. The data should have at least 3
-        dimensions, according to the 'C' memory layout, e.g. data of
+        Raw data in the time domain. The data should have at least 2
+        dimensions, with the last dimension representing the time domain
+        samples, the second to last the spherical harmonic coefficients,
+        and any leading dimensions representing optional channels. Accordingly,
+        the data should follow the 'C' memory layout, e.g. data of
         ``shape = (1, 4, 1024)`` has 1 channel with 4 spherical harmonic
-        coefficients with 1024 samples each. The data can be ``int``,
+        coefficients with 1024 samples each. When The data can be ``int``,
```
ahms5:
```suggestion
        coefficients with 1024 samples each. The data can be ``int``,
```

tests/test_spherical_harmonics_signal.py
```diff
@@ -70,9 +56,9 @@ def test_spherical_harmonic_signal_wrong_dimensions():
                      [1., 2., 3.]]).reshape(1, 5, 3)
 
     with pytest.raises(ValueError,
-                match=re.escape("Invalid number of SH channels: "
-                                f"{data.shape[-2]}. It must match "
-                                "(n_max + 1)^2.")):
+                       match=re.escape("Invalid number of spherical harmonic "
+                                       f"channels: {data.shape[-2]}. It must "
+                                       "match (n_max + 1)^2.")):
```
mberz:
```suggestion
                   match=re.escape("Invalid number of spherical harmonic "
                                   f"channels: {data.shape[-2]}. It must "
                                   "match (n_max + 1)^2.")):
```

PR #167, opened 2025-07-11
initial commit for testing common SH definitions

f-brinkmann:

Test our definition against other reference implementations. At the moment, we don't agree with all of them. Will report during the weekly. Draft for now...


ahms5 (commented):
jsut because I have just noticed it. please remove the .asv file.

mberz:
See comment above

Issue #147, opened 2025-06-03
Ruff rules are not activated yet


ahms5:
see pyproject.toml on develop.
The ruff rules are mostly commented, so we need to still fix a lot of things.

## current state (2025-07-27):

- [x] "B", # bugbear extension #148
- [x] "ARG", # Remove unused function/method arguments #149 
- [x] "E", # PEP8 errors #150 
- [x] "F", # Pyflakes #170 
- [ ] "W", # PEP8 warnings #170 
- [x] "D", # Docstring guidelines #149
- [x] "PT", # Pytest style #170 
- [ ] "A", # Avoid builtin function and type shadowing #169
- [ ] "ERA", # No commented out code #170 
- [ ] "COM", # trailing comma rules #170

f-brinkmann:
I would suggest to do this close to last before releasing 1.0 to avoid nasty conflicts with large open PRs

ahms5:
I would rather do it asap. So we can do fine tunting afterwards. otherwise things get mixed up.

PR #171, opened 2025-07-28
rename "type" parameters to avoid shadowing

### LINKED ISSUE #169: Avoid shadowing Python builtin functions
When adding more ruff checks I noted that we are shadowing `type` at three spots. I would suggest to replace `type` by `basis` in the last to cases (in analogy to the SH class) and by `matrix_type` in the first case.

```shell
spharpy/spherical.py:1056:32: A002 Function argument `type` is shadowing a Python builtin
     |
1056 | def sph_identity_matrix(n_max, type='n-nm'):
     |                                ^^^^ A002
1057 |     """Calculate a spherical harmonic identity matrix.
     |

spharpy/transforms/rotations.py:248:37: A002 Function argument `type` is shadowing a Python builtin
    |
246 |         self._n_max = value
247 | 
248 |     def as_spherical_harmonic(self, type='real'):
    |                                     ^^^^ A002
249 |         """Export the rotation operations as a spherical harmonic rotation
250 |         matrices. Supports complex and real-valued spherical harmonics.
    |

spharpy/transforms/rotations.py:284:35: A002 Function argument `type` is shadowing a Python builtin
    |
282 |         return np.squeeze(D)
283 | 
284 |     def apply(self, coefficients, type='real'):
    |                                   ^^^^ A002
285 |         """Apply the rotation to L sets of spherical harmonic coefficients.
    |
```

hoyer-a:
Closes #169 
Part of #147 

- rename `type` parameters to avoid shadowing (ruff: A)
- in case of the identity-matrix -> `matrix_type`
- when referring to sh-definition -> `basis_type`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -107,7 +107,7 @@ lint.select = [
     "NPY",    # Check all numpy related deprecations
     "D417",   # Missing argument descriptions in the docstring
     # "PT",     # Pytest style
-    # "A",      # Avoid builtin function and type shadowing
+    "A",      # Avoid builtin function and type shadowing
     # "ERA",    # No commented out code
     "NPY",    # Check all numpy related deprecations
     # "COM",    # trailing comma rules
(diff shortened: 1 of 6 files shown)


f-brinkmann (approved):

mberz (approved):
Thanks!
At least the rotation matrix changes will become obsolete though when we adapt the new SpehricalHarmonics class in the next weeks :)

PR #172, opened 2025-07-29
add new sh normalizations

tluebeck:
### Which issue(s) are closed by this pull request?

Closes #168 

### Changes proposed in this pull request:

- introduce new SH normalizations 'nm' and 'sn3d'
- update spherical.renormalization method
- update spherical_harmonic_basis, spherical_harmonic_basis_real
- update spherical_harmonic_basis_gradient, spherical_harmonic_basis_gradient_real
- update spherical_harmonics and spherial_harmonics_signal class

PR #173, opened 2025-08-08
Class containing the spherical harmonic definition

mberz:
This additional class could be used to provide the definition of the sperical harmonics to other classes.
The new class already serves as base class for the `SphericalHarmonics` class.

Requires #172 

Todo:

- [ ] Update to new monopole normalizations
- [ ] Add order (optional)

### DIFF
--- a/spharpy/classes/sh.py
+++ b/spharpy/classes/sh.py
@@ -52,11 +52,7 @@ class SphericalHarmonicDefinition:
                     "condon_shortley must be a bool or the string 'auto'")
             # If basis_type hasn't been set yet, assume "complex" by default,
             # but in practice __init__ sets basis_type before condon_shortley.
-            if self.basis_type == "complex":
-                resolved = True
-            else:
-                resolved = False
-            value = resolved
+            value = self.basis_type == "complex"
         elif not isinstance(value, bool):
             raise TypeError(
                 "condon_shortley must be a bool or the string 'auto'")
--- a/spharpy/classes/sh.py
+++ b/spharpy/classes/sh.py
@@ -91,9 +91,6 @@ class SphericalHarmonicDefinition:
             raise ValueError("Invalid channel convention, "
                              "currently only 'acn' "
                              "and 'fuma' are supported")
-        if value == "fuma" and self.n_max > 3:
-            raise ValueError("n_max > 3 is not allowed with 'fuma' " \
-                            "channel convention")
         self._channel_convention = value
 
     @property
@@ -109,9 +106,6 @@ class SphericalHarmonicDefinition:
                 "Invalid normalization, "
                 "currently only 'n3d', 'maxN', 'sn3d' are "
                 "supported")
-        if value == "maxN" and self.n_max > 3:
-            raise ValueError("n_max > 3 is not allowed with " \
-                            "'maxN' normalization")
         self._normalization = value
 
 
@@ -358,6 +352,10 @@ class SphericalHarmonics(SphericalHarmonicDefinition):
     @SphericalHarmonicDefinition.channel_convention.setter
     def channel_convention(self, value):
         """Get or set the channel order convention."""
+        if value == "fuma" and self.n_max > 3:
+            raise ValueError("n_max > 3 is not allowed with 'fuma' " \
+                            "channel convention")
+
         super(__class__, SphericalHarmonics).channel_convention.fset(
             self, value)
         if value != self._channel_convention:
@@ -366,6 +364,10 @@ class SphericalHarmonics(SphericalHarmonicDefinition):
     @SphericalHarmonicDefinition.normalization.setter
     def normalization(self, value):
         """Get or set the normalization convention."""
+        if value == "maxN" and self.n_max > 3:
+            raise ValueError(
+                "n_max > 3 is not allowed with 'maxN' normalization")
+
         super(__class__, SphericalHarmonics).normalization.fset(self, value)
         if value != self._normalization:
             self._reset_compute_attributes()
--- a/spharpy/classes/sh.py
+++ b/spharpy/classes/sh.py
@@ -28,6 +28,11 @@ class SphericalHarmonicDefinition:
             normalization="n3d",
             condon_shortley="auto",
         ):
+        self._basis_type = None
+        self._channel_convention = None
+        self._condon_shortley = None
+        self._normalization = None
+
         self.condon_shortley = condon_shortley
         self.basis_type = basis_type
         self.channel_convention = channel_convention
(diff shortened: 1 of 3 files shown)


spharpy/classes/sh.py
```diff
@@ -6,7 +6,108 @@
 import spharpy as sy
 
 
-class SphericalHarmonics:
+class SphericalHarmonicDefinition:
+    """Class storing the definition of spherical harmonics.
+
+    Attributes
+    ----------
+    condon_shortley : bool
```
f-brinkmann, line 14: Does not document `condon_shortley = 'auto'`. Maybe not needed it class would be private, but nice to have in any case.

spharpy/classes/sh.py
```diff
@@ -156,31 +257,13 @@ def __init__(
...
-                resolved = True
-            else:
-                resolved = False
-            value = resolved
-        elif not isinstance(value, bool):
-            raise TypeError(
-                "condon_shortley must be a bool or the string 'auto'")
+        super(__class__, SphericalHarmonics).condon_shortley.fset(self, value)
         if value != self._condon_shortley:
             self._reset_compute_attributes()
```
f-brinkmann, line 266: I think it must be the other way around, otherwise `value != self._condon_shortley` will always be `False`
```suggestion
        if value != self._condon_shortley:
            self._reset_compute_attributes()
            super(__class__, SphericalHarmonics).condon_shortley.fset(self, value)
```

spharpy/classes/sh.py
```diff
@@ -223,20 +306,12 @@ def coordinates(self, value):
...
+    @SphericalHarmonicDefinition.basis_type.setter
     def basis_type(self, value):
         """Get or set the type of spherical harmonic basis."""
-        if value not in ["complex", "real"]:
-            raise ValueError("Invalid basis type, only "
-                             "'complex' and 'real' are supported")
+        super(__class__, SphericalHarmonics).basis_type.fset(self, value)
         if value != self._basis_type:
             self._reset_compute_attributes()
-            self._basis_type = value
```
f-brinkmann, line 315: See above: First check old and new value, then set the new value.

spharpy/classes/sh.py
```diff
@@ -272,44 +347,28 @@ def inverse_method(self, value):
...
-                             "currently only 'acn' "
-                             "and 'fuma' are supported")
         if value == "fuma" and self.n_max > 3:
             raise ValueError("n_max > 3 is not allowed with 'fuma' " \
                             "channel convention")
+
+        super(__class__, SphericalHarmonics).channel_convention.fset(
+            self, value)
         if value != self._channel_convention:
             self._reset_compute_attributes()
```
f-brinkmann, line 360: Check first, set second

tests/test_sh_class.py
```diff
@@ -6,16 +6,101 @@
...
+from spharpy.classes.sh import SphericalHarmonicDefinition
+
+
+def test_spherical_harmonics_definition_init():
+    """Test default behavior.
+    """
+    definition = SphericalHarmonicDefinition()
+    assert definition.basis_type == 'real'
+    assert definition.normalization == 'n3d'
+    assert definition.channel_convention == 'acn'
```
f-brinkmann, line 18: add assertion for CS phase as well?

tests/test_sh_class.py
```diff
@@ -6,16 +6,101 @@
...
+
+    sph_harm = SphericalHarmonicDefinition()
+    sph_harm.basis_type = "real"
+    sph_harm.condon_shortley = "auto"
+    assert sph_harm.condon_shortley is False
+
+
+def test_setter_phase_convention_invalid():
+    sph_harm = SphericalHarmonicDefinition()
+    with pytest.raises(TypeError):
```
f-brinkmann, line 53: Add a `match=str` please. If this does not raise a ruff error, we could check for consistency of the ruff rules with respect to pyfar, where it would. @ahms5 - if I remember correctly from the weekly, you wanted to do this anyhow, right?

tests/test_sh_class.py
```diff
@@ -6,16 +6,101 @@
...
     """Test default behaviour after initialization."""
     coordinates = equiangular(n_points=8)
     sph_harm = SphericalHarmonics(n_max=2, coordinates=coordinates)
     assert sph_harm.n_max == 2
     assert np.all(sph_harm.coordinates == coordinates)
-    assert sph_harm.basis_type == 'real'
-    assert sph_harm.normalization == 'n3d'
-    assert sph_harm.channel_convention == 'acn'
     assert sph_harm.inverse_method == 'quadrature'
     assert not sph_harm.condon_shortley
```
f-brinkmann, line 105: could also be removed here and moved to the above (already commented above).

spharpy/classes/sh.py
```diff
@@ -6,7 +6,108 @@
...
+    @property
+    def condon_shortley(self):
+        """Get or set the Condon-Shortley phase term."""
+        return self._condon_shortley
+
+    @condon_shortley.setter
+    def condon_shortley(self, value):
+        """Get or set the Condon-Shortley phase term."""
+        if isinstance(value, str):
+            if value != 'auto':
```
f-brinkmann, line 52: I did a coverage report and saw that this is not tested.

spharpy/classes/sh.py
```diff
@@ -6,7 +6,108 @@
...
+        """
+        return self._basis_type
+
+    @basis_type.setter
+    def basis_type(self, value):
+        """The type of spherical harmonic basis. Can be ``'complex'`` or
+        ``'real'``.
+        """
+        if value not in ["complex", "real"]:
+            raise ValueError("Invalid basis type, only "
```
f-brinkmann, line 76: I did a coverage report and saw that this is not tested

f-brinkmann (commented):
Thanks for the effort. This will be very nice to have. Looks like many comments, but they are all rather small and and in some cases repeating.

spharpy/classes/sh.py
```diff
@@ -6,7 +6,108 @@
 import spharpy as sy
 
 
-class SphericalHarmonics:
+class SphericalHarmonicDefinition:
```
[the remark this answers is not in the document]
f-brinkmann: Yes, of course - I was only wondering about public vs. private. Public is probably not required for those examples.

tests/test_sh_class.py
mberz: I knew I'd forget one :D

mberz:
I'm currently waiting for #172 to be ready and merged to continue working on this PR.

ahms5:
it is merged now, please rerequest review if its ready

PR #162, opened 2025-06-15
add tests for docs

ahms5:
### Which issue(s) are closed by this pull request?

Closes #95

on hold: wait for #149 

### Changes proposed in this pull request:

- [x] add tests for plot module after pyfar standards
- [ ] fix tests after #149 is merged

PR #178, opened 2025-09-15
Add levels parameter to sph.plot.contour

hoyer-a:
**Add levels parameter to `sph.plot.contour` as in `contour_map`.**

Levels parameter was missing in `sph.plot.contour`.

### DIFF
--- a/spharpy/plot/spatial.py
+++ b/spharpy/plot/spatial.py
@@ -589,7 +589,7 @@ def voronoi_cells_sphere(sampling, round_decimals=13, ax=None):
     ax.set_zlabel('z[m]')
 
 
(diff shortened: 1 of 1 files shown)


f-brinkmann (changes requested):
I got an error when trying the new feature and would suggest to add a test for it
```
import spharpy as sh
import numpy as np
%matplotlib inline

sampling = sh.samplings.fliege(n_max=5)
data = np.random.rand(sampling.x.size)

sh.plot.contour(sampling, data, levels=3)
```
> TypeError: 'locator' must be an instance of matplotlib.ticker.Locator, not a int

ahms5 (changes requested):
maybe build it on top of  #162 would make things and tests easier

PR #142, opened 2025-06-01
Deprecate scipy Spherical Harmonics function

f-brinkmann:
### Changes proposed in this pull request:

closes #83 

- Use scipy's new `sph_harm_y` instead of `sph_harm` which will be deprecated in scipy 1.17.0
- Restrict scipy to version >= 1.15.0
- Add tests for `spharpy.special.spherical_harmonic` to make sure changes work as intended
- Minor corrections in the docstrings

Notes
- This requires tha latest scipy version. We should discuss if we want this. In my opinon its fine for spharpy 1.0.0.
- This removes 433073 warnings from the tests 😅

PR #180, opened 2025-09-16
Constrain scipy version to be below 1.17.0

mberz:
Required due to deprecations of special functions.

Workaround for #83

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ classifiers = [
 ]
 dependencies = [
     'numpy>=1.22',
-    'scipy',
+    'scipy<1.17.0',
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -26,7 +26,7 @@ classifiers = [
 ]
 dependencies = [
     'numpy>=1.22',
-    'scipy',
+    'scipy<1.17.0',
```
f-brinkmann, line 29: It should also be restricted to >=1.15.0 because of #142

f-brinkmann (approved):
Thanks - looks good to me.

ahms5 (approved):

PR #181, opened 2025-09-16
Workaround for the Legendre function bug in scipy

mberz:
Workaround for the bug encountered in #143 

We can merge this in case the [original issue](https://github.com/scipy/scipy/issues/23101) does not get resolved in time

### DIFF
--- a/spharpy/special.py
+++ b/spharpy/special.py
@@ -393,18 +393,18 @@ def legendre_function(n, m, z, cs_phase=True):
     z = np.atleast_1d(z)
 
     if m == 0:
-        legendre = _spspecial.eval_legendre(n, z)  # This is reliable
+        legendre = _spspecial.legendre_p(n, z)
     else:
-    # squeeze required because the legendre function introduced in scipy 1.15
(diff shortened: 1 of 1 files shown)


spharpy/special.py
```diff
@@ -392,10 +392,13 @@ def legendre_function(n, m, z, cs_phase=True):
     """
     z = np.atleast_1d(z)
 
+    if m == 0:
+        legendre = _spspecial.eval_legendre(n, z)  # This is reliable
+    else:
     # squeeze required because the legendre function introduced in scipy 1.15
     # returns a 2D array, whereas the previous function `lpmn` returned a 1D
     # array
```
ahms5, line 400:
```suggestion
        # squeeze required because the legendre function introduced in scipy 1.15
        # returns a 2D array, whereas the previous function `lpmn` returned a 1D
        # array
```

ahms5:
@sikersten 
#143 depends on scipy's fix, if they dont fix before our release, we can merge this workaround here.

PR #184, opened 2025-09-17
Add colormap encoding parameter for balloon plots

### LINKED ISSUE #174: Unexpected color encoding in balloon plots
from #162 

phase parameter depends on wheather the data is complex or not. maybe we can define 3 options 

phase=True, False, 'auto'

mberz:

Closes #174 

- Replace the phase=True/False parameter with cmap_encoding='magnitude'/'phase'


hoyer-a (approved):
Thanks, looks fine to me.

sikersten (changes requested):
Thanks, nice improvement! I only have a minor comment on the docstring, otherwise approved.

PR #187, opened 2025-09-18
Harmonize contour plots

### LINKED ISSUE #179: Harmonize contour / contour_map, fix bug
Instead of separate smaller PRs (#178, #177) it would make more sense to harmonize contour and contour_map in one branch based on #162. Like this we can adjust the tests in one go, which are parametrized for contour and contour_map anyways.

Harmonize:
- levels parameter
- passed to contour lines and filled contours
- replace pcolormesh in contour_map with contourf

Bug:
- if `levels` is an int, the colorbar fails because of `ticks=levels`. This can be fixed by creating a [MaxNLocator](https://matplotlib.org/stable/api/ticker_api.html#matplotlib.ticker.MaxNLocator) object if levels is an int. tricontour / contour does the same by default if an integer is passed as levels

hoyer-a:

Closes #179 

- Add levels parameter to `sph.plot.contour`
- Pass levels to contour-lines and filled contours
- replace `pcolormesh` with `contourf` in `sph.plot.contour_map`
- create levels for colorbar if `type(levels) is int` (bugfix)
- create copy of input data in `sph.plot.contour`, to avoid modifying the original array (bugfix)


spharpy/plot/spatial.py
```diff
@@ -952,8 +960,11 @@ def contour_map(
 
     ax.contour(xi, yi, zi, levels=levels, linewidths=0.5, colors='k',
                vmin=limits[0], vmax=limits[1], extend=extend)
-    cf = ax.pcolormesh(xi, yi, zi, cmap=cmap, shading='gouraud',
-                       vmin=limits[0], vmax=limits[1])
+    cf = ax.contourf(xi, yi, zi, levels=levels, cmap=cmap, vmin=limits[0],
+                     vmax=limits[1])
```
mberz, line 964: Actually, these can now be replaced using the private function _combined_contour, which removes the need for interpolating the data. This makes the function a lot slimmer already.

mberz:
> Not sure if using `_combined_contour` / `tricontour` is an improvement in comparison to the `_interpolate_data_on_sphere` triangulation.
> 
> At some point we have to refactor `interpolate_data_on_sphere` if we want to solve the problem with whitespace at the edges / poles.
> 
> 
> _tricontour triangulation_
> 
> 
> _manual triangulation_

I would still argue, that this is an improvement.
The reason why I'm saying this is that when using the tricontour function, the data get's used directly instead of implicitly  interpolating first. 
I really think it is beneficial to separate the visualization from the interpolation, since you cannot know what errors may arise from the interpolation. It may look adequate for the data you're showing, but there might be other cases where the provided data is sampled very sparsely, irregularly or only for a partial sphere.

f-brinkmann:
> I really think it is beneficial to separate the visualization from the interpolation
Did not follow this pull in detail. But I very much agree with this point.

PR #191, opened 2025-09-26
Phase twilight docstring not escaped math symbols

mberz:

- Converted to raw docstring

### DIFF
--- a/spharpy/plot/cmap.py
+++ b/spharpy/plot/cmap.py
@@ -5,7 +5,7 @@ import numpy as np
 
 
 def phase_twilight(lut=512):
-    """
+    r"""
     Cyclic color map for displaying phase information.
 
     This is a modified version of the twilight color map from matplotlib.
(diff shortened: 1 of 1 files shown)


spharpy/plot/cmap.py
hoyer-a, line 12:
```suggestion
    The colormap is rotated such that :math:`0` is encoded as red hues, while
```

hoyer-a, line 13:
```suggestion
    :math:`\pi` is encoded as blue hues.
```

hoyer-a (changes requested):
There is also an issue with the sphinx syntax, this should fix it.

hoyer-a (approved):
Thanks!

f-brinkmann (approved):

f-brinkmann:
Thanks. Just to check: is this going into develop because its 1.0.0 or should it go into main?

Issue #211, opened 2025-10-17
Limited parametrization of colobar properties


mberz:
The plots including a color encoding currently only allow to set if a colorbar is created or not.
This can be quite limiting. As an alternative, we could think about explicitly allow passing colorbar arguments as dictonary.
See for example this arviz function: https://python.arviz.org/en/stable/api/generated/arviz.plot_kde.html, where every `*_kwargs` gets passed to the respective function/class only.

This could give users the possibility to avoid issues such as #183.

hoyer-a:
completed in #222

PR #214, opened 2025-10-21
Plots/add utils functions

hoyer-a:
In preparation of #211 `_prepare_plot()` is added similar to `pyfar.plot._utils._prepare_plot()`: 

`_prepare_plot()`:
return figure and ax depending on ax input and projection

### DIFF
--- a/spharpy/plot/_utils.py
+++ b/spharpy/plot/_utils.py
@@ -4,42 +4,7 @@ import numpy as np
 import matplotlib.pyplot as plt
 
 
-def _add_colorbar(colorbar, fig, ax, mappable, label):
-    """
-    Add colorbar to plot.
-
-    Parameters
-    ----------
-    colorbar : bool
-        Flag indicating if a colobar should be added to the plot
-    fig : matplotlib.figure.Figure
-        Figure to plot on.
-    ax : list[matplotlib.axes.Axes]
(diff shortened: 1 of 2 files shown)


spharpy/plot/_utils.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def _prepare_plot(ax, projection=None):
+    """
+    Returns a figure to plot on.
+
+    Parameters
+    ----------
+    ax : matplotlib.axes.Axes or list, tuple or ndarray of maplotlib.axes.Axes
+        Axes to plot on. The default is None in which case the axes are
```
sikersten: Do I get it right, that this means the default is calling it as `_prepare_plot()`? Then the default for ax is missing in line 42.

spharpy/plot/_utils.py
```diff
@@ -0,0 +1,82 @@
+"""Private utility functions for plot module."""
+
+import numpy as np
+import matplotlib.pyplot as plt
+
+
+def _add_colorbar(colorbar, fig, ax, mappable, label):
```
sikersten: This function exactly matches pyfars pf.plot._utils._add_colorbar(). So why no directly use it or just wrap it here?

sikersten (changes requested):

ahms5 (approved):

PR #218, opened 2025-10-24
Abstract base class for SphericalHarmonicDefiniton and -Audio

mberz:

- Create an abstract base class for SphericalHarmonicDefinition for more flexible re-use in case n_max is not required or a getter only property
- New ABC can be used as parent in #166 
- Child classes which should not allow setting properties will need to overload the respective setter property.

### DIFF
--- a/spharpy/classes/sh.py
+++ b/spharpy/classes/sh.py
@@ -8,7 +8,13 @@ from abc import ABC, abstractmethod
 
 
 class _SphericalHarmonicBase(ABC):
-    """Base class related to spherical harmonics.
+    """Base class defining properties to parametrize spherical harmonics.
+
+    This base class serves as a base for all classes requiring a definition of
+    the spherical harmonics without explicitly setting a spherical harmonic
+    order. This class is intended for cases where the spherical harmonic order
+    is implemented as a read-only property in child classes, for example when
+    the order is implicitly defined by other parameters or inferred from data.
 
     Attributes
     ----------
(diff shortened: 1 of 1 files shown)


spharpy/classes/sh.py
```diff
@@ -158,14 +138,78 @@ def normalization(self, value):
             self._normalization = value
             self._on_property_change()
 
-    def _on_property_change(self):
+    def _on_property_change(self):  # noqa: B027
```
sikersten: Why doesn't adding `@abstractmethod` work in this case?

spharpy/classes/sh.py
```diff
@@ -4,15 +4,14 @@
 import numpy as np
 import pyfar as pf
 import spharpy as sy
+from abc import ABC, abstractmethod
 
 
-class SphericalHarmonicDefinition:
-    """Class storing the definition of spherical harmonics.
+class _SphericalHarmonicBase(ABC):
+    """Base class related to spherical harmonics.
```
sikersten: Could you please add a short explanation here similar to the one in the PR ("Abstract base class for SphericalHarmonicDefinition for more flexible re-use in case n_max is not required or a getter only property"). This would help future developers to decide whether to inherit from  `_SphericalHarmonicBase` or `SphericalHarmonicDefinition`.

spharpy/classes/sh.py
sikersten: Ok, thanks for the explanation!

sikersten (approved):

PR #227, opened 2025-11-24
Plots/Refactor contour_map plot (ax, colorbar)

hoyer-a:

merges into #222 

- return colorbar for contour_map plot
- allow input of 2 axes as tuple, list or array
- plot colorbar on second axis


tests/test_plots.py
f-brinkmann: Is the baseline only generated for this case? I assume should be genereated everywhere with something like `f'ax_option_{ax_option}'` appended to the string.

PR #229, opened 2025-12-01
Update SphericalHarmonic*Data related inits and properties

mberz:
- Removed dependency on data shape from the SphericalHarmonicAudio class since it does not contain any data related attributes
- Private functions for dimension extension and shape verification


mberz:
Closing, since manually copied into #166 by @tluebeck

PR #234, opened 2025-12-04
Fix SamplingSphere docs

f-brinkmann:

Correct wrong mentions of coordinates class and a typo.


hoyer-a (approved):
Approved, but i noticed that all the examples still use `pyfar.SamplingSphere`.
I opened an issue for that

PR #246, opened 2025-12-12
Refactoring setting inverse calculation method in SphericalHarmonics class

mberz:
There seemed to be a bug with this which breaks calculating spherical harmonics for non SamplingSphere objects.

- Allow setting the inverse method to None which is required if coordinates are passed as pf.Coordinates object.
- Simplify if-else logic for checking of valid inputs
- 'auto' now defaults to None if pf.Coordinates are passed.
- mild refactoring of formatting and import conventions:
    - changed import convention (as half of the code was already affected ;))
    - fixed indentation error 🍻

### DIFF
--- a/tests/test_sh_class.py
+++ b/tests/test_sh_class.py
@@ -3,8 +3,9 @@ Tests for spherical harmonic class.
 """
 import pytest
 import numpy as np
-from spharpy import SphericalHarmonics, SamplingSphere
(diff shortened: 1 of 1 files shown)


mberz:
> I assume the failing documentation build was caused by a bad internet connection. Docs were not changed and tests pass ob other branches.

I'll quickly restart it

tluebeck (approved):

PR #249, opened 2025-12-16
[ENH] Initialize spherical harmonics from definition

f-brinkmann:

- Add class method `SphericalHarmonics.from_definition`
- Add testing thereof

Note: Allready uses docstring suggested in #246, which will be implemented through the same pull request also here.
Requires #246


mberz (changes requested):
I'd suggest to change the order of arguments (for semantic and consistency reasons), see below)
Otherwise good to go

PR #257, opened 2025-12-22
Plots/Add style parameter to plot module

hoyer-a:
Merges into #222 

### Which issue(s) are closed by this pull request?

Closes #183 

### Changes proposed in this pull request:

- Add style parameter to plot module
- Set pyfar 'light' style as default
- Create baseline plots for light / dark pyfar style

PR #262, opened 2026-01-15
Docs/Adjust example plots for balloon and balloon_wireframe

hoyer-a:
Merges into #222 

- Adding the pyfar plot style in #257 did not improve the colorbar placement in balloon-like plots (see new baseline plots #242) 
- To get better looking examples in the docs, the data is adjusted so the balloon-data extends in x-direction

### Note

A general guide on how to generate plots without overlap issues will be added to the docs in a separate pull (#263).
### Example

| Before | After |
|--------|-------|
|  |  |


mberz (approved):
Bit of a cheeky solution, I like it :)

PR #263, opened 2026-01-15
Plots/Add general guide on the overlap issue

hoyer-a:
merges into #222 
see https://github.com/pyfar/spharpy/pull/242#discussion_r2687389549

- Add top level docs to plot module explaining the overlap issue
- Provide an example on how to individualize a plot using GridSpec and passing a list of axes to the plot function

### DIFF
--- a/spharpy/plot/__init__.py
+++ b/spharpy/plot/__init__.py
@@ -16,7 +16,7 @@ figure is :class:`matplotlib.gridspec.GridSpec`.
 
 
 Example
-~~~~~~~~
+~~~~~~~
 
 .. plot::
 


spharpy/plot/__init__.py
```diff
@@ -1,4 +1,45 @@
-"""Plotting functions for spherical data."""
+"""
+Plotting functions for spherical data.
+
+Troubleshooting - Overlapping Elements
```
f-brinkmann: Should we rephrase to something more positive?
```suggestion
Adjusting the Plot Layout
```

PR #269, opened 2026-02-06
Refactor samplings.fliege

### LINKED ISSUE #202: refactor samplings.fliege
## [spharpy.samplings.fliege](https://spharpy--190.org.readthedocs.build/en/190/modules/spharpy.samplings.html#spharpy.samplings.fliege) could use some updates in the docs, for example:
- ``n_points`` and ``n_max`` are interchangable, I would vote for removing ``n_points``
- remove the table (its just showing the equation ``n_points = (n_max + 1)**2``.) and note that order 29 is the highed possible number.
- currently the test is faling for n_max = 10, 12, 18, 20, 24, 26, 27, 29. The reason is negative weights for a few entries. If I do absolute, the weights do not sum up to 4pi. The bug must be in the [SOFiA](https://audiogroup.web.th-koeln.de/SOFiA_wiki/DOWNLOAD.html) toolbox., because they are precalculated and we just read them from disk. We decided to remove the weights.

ahms5:

Closes #202

- remove redundant ``n_points`` parameter
- remove table and add equation to doc
- simplify input check 
- split large tests into sep smaller tests
- add a test for each valid n_max. 

### open issues

currently the test is faling for n_max = 10, 12, 18, 20, 24, 26, 27, 29. The reason is negative weights for a few entries. If I do absolute, the weights do not sum up to 4pi. The bug must be in the [SOFiA](https://audiogroup.web.th-koeln.de/SOFiA_wiki/DOWNLOAD.html) toolbox., because they are precalculated and we just read them from disk.

- [ ] fix failing tests.

### DIFF
--- a/tests/test_samplings.py
+++ b/tests/test_samplings.py
@@ -471,6 +471,26 @@ def test_fliege_for_each_order(n_max):
     npt.assert_allclose(c.radius, 1, atol=1e-15)
 
 
+@pytest.mark.parametrize("n_max", [
+    1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21,
+    22, 23, 24, 25, 26, 27, 28, 29])
+def test_fliege_orthogonality(n_max):
+    """
+    Test orthogonality of the transform.
+
+    This was done after discovering https://github.com/pyfar/spharpy/issues/202
+    to make sure that the sampling can be used for SH transforms using the
+    pseudo inverse.
+    """
+
+    sampling = samplings.fliege(n_max)
+    Y = spherical_harmonic_basis_real(n_max, sampling)
+    Y_inverse = np.linalg.pinv(Y)
+
+    # if orthogonal Y_inverse @ Y must be the identity matrix
+    npt.assert_allclose(Y_inverse @ Y, np.eye((n_max + 1)**2), atol=1e-12)
+
+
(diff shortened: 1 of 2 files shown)


ahms5:
the problem is, if the weights are incorrect the points could also be incorrect. so I'll check the data from your reference and compare if you don't decide anything else in the meeting. If they lead to better results i would replace our source.

mberz:
we've decided to not return the weights for all grids which provide erroneous weights (if it occurs for one order, we remove for the entire grid.)

tests/test_samplings.py
```diff
@@ -457,35 +457,36 @@ def test_lebedev_orthogonality(degree):
...
+    c = samplings.fliege(n_max)
     assert type(c) is SamplingSphere
-    assert c.csize == 16
-
-    # test with spherical harmonic order
-    c = samplings.fliege(n_max=3)
-    assert c.csize == 16
-
-    # test default radius
+    assert c.csize == (n_max+1)**2
```
f-brinkmann: After removing the weights, we could check for orthogonality of the SH matrix as we do here:
https://github.com/pyfar/spharpy/blob/41320d331b9087f81d412cde34ecec35c366b37c/tests/test_samplings.py#L434

mberz (approved):

mberz:
Haha, nice move @f-brinkmann: Commit and self-review. That's how you guarantee that the PR you work on get's sufficient reviews :)

PR #279, opened 2026-02-12
BUG: Pass the condon_shortley phase property to the basis calculation function.

### LINKED ISSUE #281: Condon Shortley phase unused in SphericalHarmonics class
Currently the `condon_shortley` property is not passed when calling the basis matrix calculation method in the `SphericalHarmonics` class.
Seems that this was forgotten to implement or update.

mberz:

- Add the condon_shortley perperty to the basis matrix calculation method of the SphericalHarmonics class
- update the test to explicitly check if any element of the updated basis matrix is not equal to the previous matrix. Seems that the previous check could fail in the case where only the content of the array was changes, but the memory address of the (pointer of the) original array remained the same.

closes #281

### DIFF
--- a/tests/test_sh_class.py
+++ b/tests/test_sh_class.py
@@ -277,35 +277,53 @@ def test_compute_basis_caching(icosahedron_sampling):
 
     sh.n_max = n_max  # Setting to the same value should not reset cache
     # Call the method again and check that the result is the same (cache hit)
-    assert sh.basis is last_result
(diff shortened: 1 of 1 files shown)


f-brinkmann (commented):
I'm glad you noticed. That would have been a huge bug. All the added code looks good to me. I checked the tests and think we don't have a specific test for the CS pahse. I would suggest to add this here and ecplicitly test if SH matrices differ by (-1)^m.

Issue #308, opened 2026-03-30
BUG: Fliege grid files missing from binary and source distributions


mberz:
The Fliege grid relies on files which are not Python source files. These are not properly included when building the binary and source distributions.

PR #310, opened 2026-03-30
MAINT: Create CircleCI job to test against wheels (binary distributions)

### LINKED ISSUE #309: MAINT: Tests against binary distributions missing
We currently only perform testing on the source code directly.
It would be beneficial to also test against the binary distribution (i.e. the wheel which gets installed from pypi).

This can help avoiding issues such as #308

mberz:

Closes #309 

- Add test against the binary distribution (wheels) to the CircleCI pipeline
- Also re-organized the test pipeline a bit. Built products are now stored in the workspace and passed between runners to avoid duplicate steps.
- Removed commented out installation of system dependencies.

**Note:** tests are expected to fail, since this PR helps to detect issues such as reported in #308, see #309 for further info.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,20 +62,16 @@ jobs:
           command: |
             python -m build --wheel
       - run:
-          name: Install wheel in isolated venv and remove all source files from the test environment
+          name: Install wheel in isolated venv and run tests
           command: |
             python -m venv .wheeltest
             WHEEL=$(ls -1 dist/*.whl | head -n 1)
-            .wheeltest/bin/python -m pip install "$WHEEL"
-            .wheeltest/bin/python -c "import pathlib, spharpy; p = pathlib.Path(spharpy.__file__).resolve(); print(p); assert 'site-packages' in str(p), p"
+            .wheeltest/bin/python -m pip install "$WHEEL"[tests]
             TEST_PYTHON="$(pwd)/.wheeltest/bin/python"
             rm -rf /tmp/spharpy-wheel-tests
-      - run:
-          name: Run tests with wheel installation
-          command: |
             cp -r tests /tmp/spharpy-wheel-tests
             cd /tmp/spharpy-wheel-tests
-            "$TEST_PYTHON" -m pytest tests
+            "$TEST_PYTHON" -m pytest .
 
   ruff:
     parameters:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -86,8 +86,6 @@ jobs:
             # Find the wheel file in the workspace and save it to an environment variable
             WHEEL=$(ls -1 /tmp/workspace/dist/*.whl | head -n 1)
             .wheeltest/bin/python -m pip install "${WHEEL}[tests]"
-            # Create an environment variable for the path to the python executable of the isolated venv
-            TEST_PYTHON="$(pwd)/.wheeltest/bin/python"
       - run:
           name: Move test files to isolated location
           command: |
@@ -97,6 +95,8 @@ jobs:
       - run:
           name: Run tests against wheel installation (i.e. in isolated environment)
           command: |
+            # Create an environment variable for the path to the python executable of the isolated venv
+            TEST_PYTHON="$(pwd)/.wheeltest/bin/python"
             cd /tmp/spharpy-wheel-tests
             "$TEST_PYTHON" -m pytest tests
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -264,6 +264,11 @@ workflows:
           requires:
             - test_source
             - build_distribution
+          filters:
+            branches:
+              only:
+                - main
+                - develop
 
 
   test_and_publish:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -121,9 +121,6 @@ jobs:
       version: <<parameters.version>>
     steps:
       - checkout
-   #    - run:
-   #        name: Install System Dependencies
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -119,17 +159,15 @@ jobs:
...
     steps:
       - checkout
-    #  - run:
-    #      name: Install System Dependencies
-    #      command: sudo apt-get update && sudo apt-get install -y libsndfile1
+      - attach_workspace:
+          at: /tmp/workspace
       - run:
           name: install dependencies
           command: pip install ".[deploy]"
```
ahms5, line 173:
```suggestion
          command: pip install twine
```
woudnt this be sufficiant?

.circleci/config.yml
```diff
@@ -150,18 +190,17 @@ jobs:
...
```
ahms5, line 194:
```suggestion
          command: pip install twine
```
same here.

.circleci/config.yml
```diff
@@ -36,16 +59,39 @@ jobs:
...
+            python -m venv .wheeltest
+            WHEEL=$(ls -1 /tmp/workspace/dist/*.whl | head -n 1)
+            .wheeltest/bin/python -m pip install "${WHEEL}[tests]"
+            TEST_PYTHON="$(pwd)/.wheeltest/bin/python"
+            rm -rf /tmp/spharpy-wheel-tests
+            mkdir -p /tmp/spharpy-wheel-tests
+            cp -r tests /tmp/spharpy-wheel-tests/
+            cd /tmp/spharpy-wheel-tests
+            "$TEST_PYTHON" -m pytest tests
+
```
ahms5: this is not that easy to read, maybe we can spit this into 3 blocks, sth like create environment, move files, and test. In this way the maintainance would be easier.

ahms5 (changes requested):
I love it. I wonder if we really need to ``test_source``, as ``test_wheel`` should also fail in case ``test_source`` fails.

.circleci/config.yml
mberz: In theory yes, but I'd like to avoid managing dependencies in the CircleCI config and the pyproject.toml. This way it's always sufficient to extend and update in a single place.

.circleci/config.yml
mberz: see comment above

.circleci/config.yml
mberz: I think it's plenty easy to read 💩 

... I'll update :)

mberz:
> I love it. I wonder if we really need to `test_source`, as `test_wheel` should also fail in case `test_source` fails.

In general I agree, my first draft of this did exactly that. 
In the end I extended to both so that it's easier to spot where issues are coming from. But you're right, in probably 95% of the cases there's not really a difference. It's more the 5% might be harder to figure out why they actually fail.

.circleci/config.yml
ahms5: I see your point. My thinking here is that this job is only invoking twine, which doesn’t depend on our project’s dependencies or anything that could vary with the package—it’s just a standalone tool.
We could compare it to ruff, but in that case we pin a specific version and install it via the project configuration. Imo this makes sence here, but we dont have this requriedment for twine.
If we want to change the dependency of twine we also need to change this workflow here.
Right now, installing all packages takes about 29 seconds in CI. Installing only twine should be significantly faster. This also scales better, since this job will run for every package on each pull request.

.circleci/config.yml
```diff
@@ -142,6 +180,8 @@ jobs:
       version: <<parameters.version>>
     steps:
       - checkout
+      - attach_workspace:
+          at: /tmp/workspace
       - run:
           name: Install System Dependencies
           command: sudo apt-get update && sudo apt-get install -y libsndfile1
```
ahms5:
```suggestion
```

I think this is nor required here, isnt it?

.circleci/config.yml
```diff
@@ -170,44 +209,64 @@ workflows:
...
           matrix:
             parameters:
               version:
                 - "3.14"
           requires:
-            - build_and_test
+            - test_source
+
+      - test_wheel:
+          matrix:
```
ahms5, line 233: to reduce redandency we could also jsut run test_wheel on main and develop and not for each pull in a pr

.circleci/config.yml
mberz: Sorry, I still don't really get the point. You could also pin to a specific version within the CircleCI config. My point is more about central management of dependencies in the toml, instead of individual management of dependencies in different places.

I realized though, that its probably more useful to check the binary distribution within the built job instead of creating a new job. This probably saves more time than the 29 secs you mentioned.

mberz: To be honest, I think this job can be removed altogether.
It only checks if the meta-data provided in the pyproject.toml is according to what PyPI expects. More specifically, it only checks the content of the item `long_description`.
I think it's sufficient to check this before truly releasing a new version and remove the check in PRs, etc.

mberz: https://twine.readthedocs.io/en/stable/internal/twine.commands.check.html#twine.commands.check.check

.circleci/config.yml
mberz: I'll update that at the end, to make sure it still triggers as long as work on this PR is ongoing

.circleci/config.yml
ahms5: that would also work for me

ahms5 (approved):

PR #314, opened 2026-04-02
WIP: Check if the test can be run without a new environment

mberz:


mberz:
Not working as intended, closing

PR #327, opened 2026-04-25
Introduce property sh_caxis

### LINKED ISSUE #326: FEAT: Create a spherical harmonic dimension/axis property
We decided to add a property to the SphericalHarmonicSignal which specifies the dimension in which the spherical harmonic coefficients are stored. 
This relaxes the limitation that the spherical harmonic coefficients are limited to the last dimension in the cshape (i.e. the last dimension of the underlying data array)
Additionally it allows storing spherical harmonic coefficients in multiple axes (MIMO data).
Indexing is most likely only feasible by referencing from the last axis of the data array, since operations like band pass filtering always prepend axes.

Required adaptions are:
- SphericalHarmonicRotation
- the spherical harmonic transform (in development)

tluebeck:

Closes #326 

- Introduce property `caxis_spherical_harmonics` in `_SphericalHarmonicsAudio`, `SphericalHarmonicTimeData` , `SphericalHarmonicFrequencyData`, and `SphericalHarmonicSignal` which defines the axis along which SHSignals store SH coefficients
- Adapt rotation accordingly

### DIFF
--- a/spharpy/classes/audio.py
+++ b/spharpy/classes/audio.py
@@ -63,7 +63,8 @@ def _assert_valid_number_of_sh_channels(shape, sh_axis):
     shape : tuple, int
         Shape of the data array.
     sh_axis : int
-        Axis which stores the spherical harmonic coefficients
+        Specifies which axis of data holds the spherical harmonic coefficients.
+        Negative indexing, i.e., interpreted relative to the end of the array.
 
     Raises
     ------
@@ -98,6 +99,9 @@ def _convert_to_standard_definition(
     channel_convention : str
         Channel ordering convention, either ``'ACN'`` or ``'FuMa'``.
         (FuMa is only supported up to 3rd order)
+    sh_axis : int
+        Specifies which axis of data holds the spherical harmonic coefficients.
+        Negative indexing, i.e., interpreted relative to the end of the array.
 
     Returns
     -------
@@ -133,6 +137,9 @@ def _convert_from_standard_definition(
     channel_convention : str
         Channel ordering convention, either ``'ACN'`` or ``'FuMa'``.
         (FuMa is only supported up to 3rd order)
+    sh_axis : int
+        Specifies which axis of data holds the spherical harmonic coefficients.
+        Negative indexing, i.e., interpreted relative to the end of the array.
 
     Returns
     -------
@@ -186,6 +193,9 @@ class _SphericalHarmonicAudio(_Audio, _SphericalHarmonicBase, ABC):
         Domain of data. The default is ``'time'``
     comment : str
         A comment related to `data`. The default is ``None``.
+    sh_caxis : int
+        Specifies which axis of data holds the spherical harmonic coefficients.
+        Negative indexing, i.e., interpreted relative to the end of the array.
 
     """
 
@@ -270,6 +280,10 @@ class SphericalHarmonicTimeData(_SphericalHarmonicAudio, TimeData):
     is_complex : bool, optional
         A flag which indicates if the time data are real or complex-valued.
         The default is ``False``.
+    sh_caxis : int
+        Specifies which axis of data holds the spherical harmonic coefficients.
+        Negative indexing, i.e., interpreted relative to the end of the array.
+        default is second last axis (-2).
     """
 
     def __init__(self, data, times, basis_type, normalization,
@@ -322,6 +336,10 @@ class SphericalHarmonicTimeData(_SphericalHarmonicAudio, TimeData):
         is_complex : bool, optional
             A flag which indicates if the time data are real or complex-valued.
             The default is ``False``.
+        sh_caxis : int
+            Specifies which axis of data holds the spherical harmonic
+            coefficients. Negative indexing, i.e., interpreted relative to the
+            end of the array. The default is second last axis (-2).
         """
         return cls(data, times,
                    basis_type=sh_definition.basis_type,
@@ -388,6 +406,10 @@ class SphericalHarmonicFrequencyData(_SphericalHarmonicAudio, FrequencyData):
         and ``False`` for real `basis_type`.
     comment : str
         A comment related to `data`. The default is ``""``.
+    sh_caxis : int
+        Specifies which axis of data holds the spherical harmonic coefficients.
+        Negative indexing, i.e., interpreted relative to the end of the array.
+        The default is second last axis (-2).
     """
 
     def __init__(self, data, frequencies, basis_type, normalization,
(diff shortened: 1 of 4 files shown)


f-brinkmann (changes requested):
@tluebeck I think we have to differentiate between `caxis` and `axis`. The SH audio objects should deal with the `sh_caxis` and the default should be `-1`. Functions that work on the SH data (numpy array) directly, such as `_assert_valud_number_of_sh_channels` must get the `sh_axis = sh_caxis - 1`. Do you agree or did I miss something?

At some point, we should add a check if `sh_caxis<0` and `np.abs(sh_caxis)<data.cdim`

tluebeck:
I agree, but I'm not sure if we want to pass "sh_axis" to SHAudio object's inits and internally recalculate to sh_caxis or if we should let the user pass "sh_caxis" directly.

tluebeck:
I decided to pass sh_caxis directly.

spharpy/classes/audio.py
```diff
@@ -281,22 +282,28 @@ class SphericalHarmonicTimeData(_SphericalHarmonicAudio, TimeData):
         A flag which indicates if the time data are real or complex-valued.
         The default is ``False``.
     sh_caxis : int
-        Specifies which axis of data holds the spherical harmonic coefficients.
-        Negative indexing, i.e., interpreted relative to the end of the array.
-        default is second last axis (-2).
+        Specifies which channel axis of data holds the spherical harmonic
```
f-brinkmann, line 290: We could additionally link

https://pyfar-gallery.readthedocs.io/en/latest/gallery/interactive/pyfar_audio_objects.html#Signal-cshape,-length,-and-caxis

to geive background on the caxis conept, as we do in the pyfar documentaiton.

spharpy/classes/audio.py
```diff
@@ -281,22 +282,28 @@ class SphericalHarmonicTimeData(_SphericalHarmonicAudio, TimeData):
...
+        coefficients. Negative indexing, i.e., interpreted relative to the end
```
f-brinkmann: Could be more clear, that this is required. maybe:

'Must be an integer smaller than or equal to ``-1``. The default ``-1`` referrs to the last channel axis and a value of ``-2`` would refer to the second last channel axis.'

spharpy/classes/audio.py
```diff
@@ -281,22 +282,28 @@ class SphericalHarmonicTimeData(_SphericalHarmonicAudio, TimeData):
...
             raise ValueError(
                 "Complex spherical harmonic basis requires "
-                "complex time data. Set is_complex=True.")
+                "complex time data. Set is_complex=True.")        
+
+        if sh_caxis > 0:
+            raise ValueError("sh_caxis has to be a negative integer.")
+        if abs(sh_caxis) >= data.ndim:
+            raise ValueError(f"sh_caxis ({sh_caxis}) exceeds the number of "
+                             f"dimensions of data ({data.ndim})")
```
f-brinkmann: Should we move the `sh_caxis` checks to the end of the initialization` We could then use `abs(sh_caxis) >= self.cdim` otherwise I think we also have to test `abs(sh_caxis) >= data.ndim + 1`.

spharpy/classes/audio.py
```diff
@@ -574,10 +587,16 @@ def __init__(self,
...
                  comment="",
                  is_complex=False,
-                 sh_caxis=-2):
+                 sh_caxis=-1):
+
+        if sh_caxis > 0:
+            raise ValueError("sh_caxis has to be a negative integer.")
+        if abs(sh_caxis) >= data.ndim:
+            raise ValueError(f"sh_caxis ({sh_caxis}) exceeds the number of "
+                             f"dimensions of data ({data.ndim})")
```
f-brinkmann: See comment above (`cdim` or `ndim + 1`)

f-brinkmann (changes requested):
Thanks for changing. I think I found one thing that needs further adjustment and had only comments on the docs apart from that.

spharpy/classes/audio.py
tluebeck: We should check that before `data = _atleast_3d_first_dimension(data)` and `_assert_valid_number_of_sh_channels(data.shape, sh_caxis-1)`, otherwise it crashes.

tests/test_spherical_harmonics_audio.py
```diff
@@ -6,6 +6,7 @@
 import numpy as np
 import numpy.testing as npt
 import pytest
+import re
```
mberz, line 9: probably obsolete?

spharpy/classes/audio.py
```diff
@@ -195,11 +213,18 @@ def __init__(self, basis_type, normalization, channel_convention,
...
+
     @property
     def n_max(self):
         """Get or set the spherical harmonic order."""
         return int(np.sqrt(self.cshape[-1])-1)
 
+    @property
+    def sh_caxis(self):
+        """Get the spherical harmonic axis"""
+        return self._sh_caxis
```
mberz:
```suggestion
    def caxis_spherical_harmonics(self):
        """Get the spherical harmonic axis"""
        return self._sh_caxis
```

I've thought about it a bit and came to the conclusion that I'd prefer a naming convention where additional info is appended instead of prepended. In my opinion that makes things a bit clearer. (also related concepts are closed together in terms of auto-complete, documentation, etc.

spharpy/classes/audio.py
```diff
@@ -259,30 +286,44 @@ class SphericalHarmonicTimeData(_SphericalHarmonicAudio, TimeData):
...
-                 is_complex=False):
+                 is_complex=False, sh_caxis=-1):
 
         if not is_complex and basis_type == 'complex':
             raise ValueError(
                 "Complex spherical harmonic basis requires "
                 "complex time data. Set is_complex=True.")
 
+        if sh_caxis > 0:
+            raise ValueError("sh_caxis has to be a negative integer.")
```
mberz: I don't think that's a necessary condition. It's just a good default to start from the end.

mberz (changes requested):
I've added some comments below.

Additionally, I have another comment:
In one of the last meetings, we discussed support for multiple dimensions containing spherical harmonic coefficients. In this implementation this is not possible, correct?
Or it simply would not be covered, but users would need to add the data in an additional dimension which is unaffected by all checks.
Would you think it's feasible to implement this as well?

spharpy/classes/audio.py
tluebeck: Sounds reasonable to me.

tluebeck:
Close #327 in favor of #329

PR #329, opened 2026-05-18
Feature/add sh axis property multichannel

### LINKED ISSUE #326: FEAT: Create a spherical harmonic dimension/axis property

tluebeck:

Closes #326

- Introduce property `caxis_spherical_harmonics` in `_SphericalHarmonicsAudio`, `SphericalHarmonicTimeData` , `SphericalHarmonicFrequencyData`, and `SphericalHarmonicSignal` which defines the axis along which SHSignals  store SH coefficients
- Adapt rotation accordingly
- allow `caxis_spherical_harmonics` property for multichannel SH data

### DIFF
--- a/spharpy/classes/audio.py
+++ b/spharpy/classes/audio.py
@@ -63,7 +63,8 @@ def _assert_valid_number_of_sh_channels(shape, sh_axis):
     shape : tuple, int
         Shape of the data array.
     sh_axis : int
-        Axis which stores the spherical harmonic coefficients
+        Specifies which axis of data holds the spherical harmonic coefficients.
+        Negative indexing, i.e., interpreted relative to the end of the array.
 
     Raises
     ------
@@ -98,6 +99,9 @@ def _convert_to_standard_definition(
     channel_convention : str
         Channel ordering convention, either ``'ACN'`` or ``'FuMa'``.
         (FuMa is only supported up to 3rd order)
+    sh_axis : int
+        Specifies which axis of data holds the spherical harmonic coefficients.
+        Negative indexing, i.e., interpreted relative to the end of the array.
 
     Returns
     -------
@@ -133,6 +137,9 @@ def _convert_from_standard_definition(
     channel_convention : str
         Channel ordering convention, either ``'ACN'`` or ``'FuMa'``.
         (FuMa is only supported up to 3rd order)
+    sh_axis : int
+        Specifies which axis of data holds the spherical harmonic coefficients.
+        Negative indexing, i.e., interpreted relative to the end of the array.
 
     Returns
     -------
@@ -186,6 +193,9 @@ class _SphericalHarmonicAudio(_Audio, _SphericalHarmonicBase, ABC):
         Domain of data. The default is ``'time'``
     comment : str
         A comment related to `data`. The default is ``None``.
+    sh_caxis : int
+        Specifies which axis of data holds the spherical harmonic coefficients.
+        Negative indexing, i.e., interpreted relative to the end of the array.
 
     """
 
@@ -270,6 +280,10 @@ class SphericalHarmonicTimeData(_SphericalHarmonicAudio, TimeData):
     is_complex : bool, optional
         A flag which indicates if the time data are real or complex-valued.
         The default is ``False``.
+    sh_caxis : int
+        Specifies which axis of data holds the spherical harmonic coefficients.
+        Negative indexing, i.e., interpreted relative to the end of the array.
+        default is second last axis (-2).
     """
 
     def __init__(self, data, times, basis_type, normalization,
@@ -322,6 +336,10 @@ class SphericalHarmonicTimeData(_SphericalHarmonicAudio, TimeData):
         is_complex : bool, optional
             A flag which indicates if the time data are real or complex-valued.
             The default is ``False``.
+        sh_caxis : int
+            Specifies which axis of data holds the spherical harmonic
+            coefficients. Negative indexing, i.e., interpreted relative to the
+            end of the array. The default is second last axis (-2).
         """
         return cls(data, times,
                    basis_type=sh_definition.basis_type,
@@ -388,6 +406,10 @@ class SphericalHarmonicFrequencyData(_SphericalHarmonicAudio, FrequencyData):
         and ``False`` for real `basis_type`.
     comment : str
         A comment related to `data`. The default is ``""``.
+    sh_caxis : int
+        Specifies which axis of data holds the spherical harmonic coefficients.
+        Negative indexing, i.e., interpreted relative to the end of the array.
+        The default is second last axis (-2).
     """
 
     def __init__(self, data, frequencies, basis_type, normalization,
@@ -430,6 +452,10 @@ class SphericalHarmonicFrequencyData(_SphericalHarmonicAudio, FrequencyData):
             the size of the last dimension of `data`, i.e., ``data.shape[-1]``.
         comment : str
             A comment related to `data`. The default is ``None``.
+        sh_caxis : int
+            Specifies which axis of data holds the spherical harmonic
+            coefficients. Negative indexing, i.e., interpreted relative to the
+            end of the array. The default is second last axis (-2).
         """
         return cls(data, frequencies,
                    basis_type=sh_definition.basis_type,
@@ -520,6 +546,10 @@ class SphericalHarmonicSignal(_SphericalHarmonicAudio, Signal):
         Specifies if the underlying time domain data are complex
         or real-valued. If ``True`` and `domain` is ``'time'``, the
         input data will be cast to complex. The default is ``False``.
+    sh_caxis : int
+        Specifies which axis of data holds the spherical harmonic coefficients.
+        Negative indexing, i.e., interpreted relative to the end of the array.
+        The default is second last axis (-2).
 
     References
     ----------
@@ -599,6 +629,10 @@ class SphericalHarmonicSignal(_SphericalHarmonicAudio, Signal):
             Specifies if the underlying time domain data are complex
             or real-valued. If ``True`` and `domain` is ``'time'``, the
             input data will be cast to complex. The default is ``False``.
+        sh_caxis : int
+            Specifies which axis of data holds the spherical harmonic
+            coefficients. Negative indexing, i.e., interpreted relative to the
+            end of the array. The default is second last axis (-2).
         """
         return cls(data, sampling_rate,
                    basis_type=sh_definition.basis_type,
--- a/spharpy/classes/audio.py
+++ b/spharpy/classes/audio.py
@@ -299,7 +299,7 @@ class SphericalHarmonicTimeData(_SphericalHarmonicAudio, TimeData):
         _assert_valid_number_of_sh_channels(data.shape, sh_caxis)
 
         if sh_caxis > 0:
-            raise AttributeError("sh_caxis has to be a negative integer.")
+            raise ValueError("sh_caxis has to be a negative integer.")
 
         _SphericalHarmonicAudio.__init__(
             self, basis_type, normalization, channel_convention,
@@ -422,7 +422,7 @@ class SphericalHarmonicFrequencyData(_SphericalHarmonicAudio, FrequencyData):
         _assert_valid_number_of_sh_channels(data.shape, sh_caxis)
 
         if sh_caxis > 0:
-            raise AttributeError("sh_caxis has to be a negative integer.")
+            raise ValueError("sh_caxis has to be a negative integer.")
 
         _SphericalHarmonicAudio.__init__(
             self, basis_type, normalization, channel_convention,
@@ -586,7 +586,7 @@ class SphericalHarmonicSignal(_SphericalHarmonicAudio, Signal):
         _assert_valid_number_of_sh_channels(data.shape, sh_caxis)
 
         if sh_caxis > 0:
-            raise AttributeError("sh_caxis has to be a negative integer.")
+            raise ValueError("sh_caxis has to be a negative integer.")
 
         _SphericalHarmonicAudio.__init__(
             self, basis_type, normalization, channel_convention,
--- a/spharpy/classes/audio.py
+++ b/spharpy/classes/audio.py
@@ -194,8 +194,9 @@ class _SphericalHarmonicAudio(_Audio, _SphericalHarmonicBase, ABC):
     comment : str
         A comment related to `data`. The default is ``None``.
     sh_caxis : int
-        Specifies which axis of data holds the spherical harmonic coefficients.
-        Negative indexing, i.e., interpreted relative to the end of the array.
+        Specifies which channel axis of data holds the spherical harmonic
+        coefficients. Negative indexing, i.e., interpreted relative to the end
+        of the array.
 
     """
 
@@ -281,14 +282,14 @@ class SphericalHarmonicTimeData(_SphericalHarmonicAudio, TimeData):
         A flag which indicates if the time data are real or complex-valued.
         The default is ``False``.
     sh_caxis : int
-        Specifies which axis of data holds the spherical harmonic coefficients.
-        Negative indexing, i.e., interpreted relative to the end of the array.
-        default is second last axis (-2).
+        Specifies which channel axis of data holds the spherical harmonic
+        coefficients. Negative indexing, i.e., interpreted relative to the end
+        of the array. The default is -1.
     """
 
     def __init__(self, data, times, basis_type, normalization,
                  channel_convention, condon_shortley, comment="",
-                 is_complex=False, sh_caxis=-2):
+                 is_complex=False, sh_caxis=-1):
 
         if not is_complex and basis_type == 'complex':
             raise ValueError(
@@ -296,7 +297,7 @@ class SphericalHarmonicTimeData(_SphericalHarmonicAudio, TimeData):
                 "complex time data. Set is_complex=True.")
 
         data = _atleast_3d_first_dimension(data)
-        _assert_valid_number_of_sh_channels(data.shape, sh_caxis)
+        _assert_valid_number_of_sh_channels(data.shape, sh_caxis-1)
 
         if sh_caxis > 0:
             raise ValueError("sh_caxis has to be a negative integer.")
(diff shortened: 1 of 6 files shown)


spharpy/spherical.py
```diff
@@ -282,7 +282,20 @@ def renormalize(data, channel_convention, current_norm, target_norm, axis):
...
+        axis = (axis,)
+
+    # check if all sh_channels match
+    sh_channels = [data.shape[a] for a in axis]
+
+    if len(set(sh_channels)) != 1:
+        raise ValueError(
+            "All SH axes must have the same number of channels, "
+            f"but got {sh_channels}"
+        )
```
mberz: I think this check should not be necessary, since the order governs the total number of channels. So all axes containing spherical harmonic data should have the same length. 

So this check should probably be moved to the respective SphericalHarmonic* classes.

spharpy/spherical.py
```diff
@@ -375,15 +392,32 @@ def change_channel_convention(data, current, target, axis):
     if current == target:
         return data
 
-    acn = np.arange(data.shape[axis])
+    # check if all sh_channels match
+    sh_channels = [data.shape[a] for a in axis]
+
+    if len(set(sh_channels)) != 1:
```
mberz: same comment as above. Implementing the check also makes individual checks in each function obsolete.

mberz (commented):
Looks very much feasible!
I think we'll need an update to the n_max property and additional checks to ensure that the order of each dimensions corresponding to spherical harmonic data matches.
Or would you suggest to allow different orders for each dimension?

spharpy/classes/audio.py
```diff
@@ -228,13 +268,15 @@ class SphericalHarmonicTimeData(_SphericalHarmonicAudio, TimeData):
     data : array, double
         Raw data in the time domain. The data should have at least 2
         dimensions, with the last dimension representing the time domain
-        samples, the second to last the spherical harmonic coefficients,
-        and any leading dimensions representing optional channels. Accordingly,
-        the data should follow the 'C' memory layout, e.g. data of
+        samples and the second to last the spherical harmonic coefficients. If
+        the raw data have more then 3 dimensions `caxis_spherical_harmonics`
+        defines the axis holding the spherical harmonic coefficients. The
+        default is -1 (second to last channel axis). Accordingly, the default
```
f-brinkmann: should be last caxis or second to last axis
```suggestion
        default is -1 (last channel axis). Accordingly, the default
```
(should be changed everywhere)

spharpy/classes/audio.py
```diff
@@ -228,13 +268,15 @@ class SphericalHarmonicTimeData(_SphericalHarmonicAudio, TimeData):
...
+        data shape follows the 'C' memory layout, e.g. data of
         ``shape = (1, 4, 1024)`` has 1 channel with 4 spherical harmonic
-        coefficients with 1024 samples each. The data can be ``int``,
-        ``float`` or ``complex``. Data of type ``int`` is converted to
-        ``float``.
+        coefficients with 1024 samples each, and
+        `caxis_spherical_harmonics` = -1. The data can be ``int``, ``float``
```
f-brinkmann:
```suggestion
        a `caxis_spherical_harmonics` of -1. The data can be ``int``, ``float``
```

or

```suggestion
        ``caxis_spherical_harmonics = -1``. The data can be ``int``, ``float``
```

(should be changed everywhere)

spharpy/classes/audio.py
```diff
@@ -259,30 +301,57 @@ class SphericalHarmonicTimeData(_SphericalHarmonicAudio, TimeData):
...
+                 is_complex=False, caxis_spherical_harmonics=-1):
 
         if not is_complex and basis_type == 'complex':
             raise ValueError(
                 "Complex spherical harmonic basis requires "
                 "complex time data. Set is_complex=True.")
+        if abs(caxis_spherical_harmonics) > data.ndim:
+            raise ValueError(
+                    f"caxis_spherical_harmonics ({caxis_spherical_harmonics}) "
+                    f"exceeds the number of dimensions of data ({data.ndim})")
```
f-brinkmann: I think this fails if caxis_spherical_harmonics is a tuple. It can probably be omitted because it is checked below as well.

spharpy/classes/audio.py
```diff
@@ -259,30 +301,57 @@ class SphericalHarmonicTimeData(_SphericalHarmonicAudio, TimeData):
...
+
+        if isinstance(caxis_spherical_harmonics, int):
+            caxis_spherical_harmonics = (caxis_spherical_harmonics, )
+
+        for caxis in caxis_spherical_harmonics:
+            if abs(caxis) > data.ndim:
+                raise ValueError(
+                    f"caxis_spherical_harmonics "
+                    f"({caxis}) exceeds the number "
+                    f"of dimensions of data ({data.ndim})")
```
f-brinkmann: Isn't this code identical for all three SHAudio classes? I would suggest to put it into a private function to make it reusable

spharpy/classes/audio.py
```diff
@@ -259,30 +301,57 @@ class SphericalHarmonicTimeData(_SphericalHarmonicAudio, TimeData):
...
+        if abs(caxis_spherical_harmonics) > data.ndim:
+            raise ValueError(
+                    f"caxis_spherical_harmonics ({caxis_spherical_harmonics}) "
+                    f"exceeds the number of dimensions of data ({data.ndim})")
+
+        if isinstance(caxis_spherical_harmonics, int):
+            caxis_spherical_harmonics = (caxis_spherical_harmonics, )
+
+        for caxis in caxis_spherical_harmonics:
+            if abs(caxis) > data.ndim:
```
f-brinkmann: Wouldn't this fail, if data is an list but not an ndarray? I think this should be moved after `data = _atleast_3d_first_dimension(data)`

spharpy/classes/audio.py
```diff
@@ -259,30 +301,57 @@ class SphericalHarmonicTimeData(_SphericalHarmonicAudio, TimeData):
...
+                raise ValueError(
+                    f"caxis_spherical_harmonics "
+                    f"({caxis}) exceeds the number "
+                    f"of dimensions of data ({data.ndim})")
 
         data = _atleast_3d_first_dimension(data)
-        _assert_valid_number_of_sh_channels(data.shape)
+
+        axes_sh = tuple(x - 1 for x in caxis_spherical_harmonics)
```
f-brinkmann: This line repeats very often. Wouldn't it be better to move this into the function and directly pass caxis_spherical_harmonics instead?

spharpy/classes/audio.py
```diff
@@ -259,30 +301,57 @@ class SphericalHarmonicTimeData(_SphericalHarmonicAudio, TimeData):
     is_complex : bool, optional
         A flag which indicates if the time data are real or complex-valued.
         The default is ``False``.
+    caxis_spherical_harmonics : int, tuple
```
f-brinkmann, line 455: I think the current implementation misses a test if all elements are int in case a tuple is passed.

spharpy/transforms/rotations.py
```diff
@@ -288,7 +288,18 @@ def apply(
         D = self.as_spherical_harmonic_matrix(target_definition)
         M = np.linalg.multi_dot(list(D)) if D.ndim > 2 else D
 
-        rotated_data = np.einsum('ij, ljt -> lit', M, target._data)
+        data = target._data
+        sh_caxis = target.caxis_spherical_harmonics - 1
```
f-brinkmann: I think this fails if caxis_spherical_harmonics is a tuple

spharpy/classes/audio.py
```diff
@@ -195,11 +224,22 @@ def __init__(self, basis_type, normalization, channel_convention,
...
     def n_max(self):
         """Get or set the spherical harmonic order."""
         return int(np.sqrt(self.cshape[-1])-1)
 
+    @property
+    def caxis_spherical_harmonics(self):
+        """Get the spherical harmonic axis."""
+
+        if len(self._caxis_spherical_harmonics) == 1:
+            return self._caxis_spherical_harmonics[0]
```
f-brinkmann: Returning either an int or a tuple might make things hard elsewhere. I found a potential bug in rotation due to this. I would suggest to always return the tuple.

spharpy/transforms/rotations.py
```diff
@@ -288,7 +288,18 @@ def apply(
         D = self.as_spherical_harmonic_matrix(target_definition)
         M = np.linalg.multi_dot(list(D)) if D.ndim > 2 else D
 
-        rotated_data = np.einsum('ij, ljt -> lit', M, target._data)
+        data = target._data
```
f-brinkmann: This is a mutable, which has the danger of changing things in target if not being careful. Because it is only used once, I would suggest to remove this here and use target._data instead of data below.

spharpy/transforms/rotations.py
```diff
@@ -288,7 +288,18 @@ def apply(
...
+        sh_caxis = target.caxis_spherical_harmonics - 1
+
+        # move SH axis to front
+        data = np.moveaxis(data, sh_caxis, 0)
```
f-brinkmann: The numpy docs say that source and destination axes must be unique, so I assume this must be

```suggestion
        data = np.moveaxis(data, sh_caxis, np.arange(len(sh_caxis)))
```

for things to work if we a have a tuple of SH caxes

spharpy/transforms/rotations.py
```diff
@@ -288,7 +288,18 @@ def apply(
...
+
+        # apply rotation
+        rotated_data = np.tensordot(M, data, axes=(1, 0))
+
+        # move SH axis back to original position
+        rotated_data = np.moveaxis(rotated_data, 0, sh_caxis)
```
f-brinkmann: see above

spharpy/spherical.py
```diff
@@ -282,10 +282,14 @@ def renormalize(data, channel_convention, current_norm, target_norm, axis):
     data : ndarray
         Renormalized data
     """
-    sh_channels = data.shape[axis]
+    if isinstance(axis, int):
+        axis = (axis,)
```
f-brinkmann, line 286: Would not be required anymore if axis was always a tuple.

spharpy/spherical.py
```diff
@@ -314,10 +318,11 @@ def renormalize(data, channel_convention, current_norm, target_norm, axis):
     # sure that the factors can be applied, new axes must be added. This is
     # done by reshaping to the following shape
     shape = [1] * data.ndim
-    shape[axis] = data.shape[axis]
+
+    shape[axis[0]] = sh_channels
```
f-brinkmann: I think this does not work as intended yet. I get the same result for renormalizing an array with a single and two SH axes:

```python
import spharpy
import numpy as np

data = np.ones((1, 4, 4, 8))
channel_convention = 'ACN'
current_norm = 'N3D'
target_norm = 'SN3D'

data1 = spharpy.spherical.renormalize(
    data, channel_convention, current_norm, target_norm, (1, ))
data2 = spharpy.spherical.renormalize(
    data, channel_convention, current_norm, target_norm, (1, 2))

np.all(data1 == data2)  # is True but should not be true (?)
```

spharpy/spherical.py
```diff
@@ -364,6 +369,9 @@ def change_channel_convention(data, current, target, axis):
     data : ndarray
         Data with changed channel convention
     """
+    if isinstance(axis, int):
+        axis = (axis,)
```
f-brinkmann, line 377: Would not be required anymore of axis was always a tuple.

f-brinkmann (changes requested):
Thanks for pushing ths forward. Docs look mainly good but I think I found one bug and have three things for general discussion. I did not yet review the tests.

1. I think renormalize does not work as intended yet (see detailed comment). If true, tests should be adapted to detect the unwanted behavior and test the desired.
2. I think it would be more consistent and easier to handle if `caxis_spherical_harmonics` would always be a tuple.
3. This is the biggest: We are inheriting from `pyfar._Audio` which has the `reshape`, `transpose` and `T` methods that move around axes. I think we need to overload them in `_SphericalHarmonicAudio` to change `caxis_spherical_harmonics` accordingly. Should be rather straigt forward: call the parent function first, then update `caxis_spherical_harmonics`. For saftey I would suggest to check dimensions again and would suggest to add a private setter for `caxis_spherical_harmonics` to `_SphericalHarmonicAudio` that does the checks. This would also reduce code redundancy, because these checks are implemented three times at the moment (once in each SH Audio class).

spharpy/spherical.py
tluebeck: I changed it for all the private methods, but I think for convenience, it would be nice to allow both, int and tuple here for the public methods.

tluebeck:
Thanks for the review. Good catches. I think everything is adapted and fixed now. Maybe we should discuss on the new behavior of reshaping SHsignals. Do we want to reshape caxis_spherical_harmonics as well? I think that wouldn't make sense.

f-brinkmann:
In addition to the above, I noticed that `_Audio.flatten()` might also be difficult because it would interleave SH channels with non SH channels. Running an LLM on all our repos found the following uses on a Signal level:

- all pyfar.plot functions flatten the signal
- pyfar/dsp/filter/gammatone.py:391 and pyfar/dsp/interpolation.py:801 and some pyrato.edc methods flatten the input and reshape after processing

All usages are fine since `_Audio.flatten()` returns a copy of the Signal. Because all functions could be usefull with an SH Signal, how can we make this work?

- `SHAudio.flatten` could return an Audio object (not SHAudio)
- gammatone and interpolation would return an SHAudio object (fine)
- EDC methods return TimeData and information about the SH axes will be lost (not ideal)

spharpy/classes/audio.py
```diff
@@ -55,14 +55,46 @@ def _atleast_3d_first_dimension(data):
     return data[np.newaxis, ...] if data.ndim < 3 else data
 
 
-def _assert_valid_number_of_sh_channels(shape, sh_axis):
+def _assert_valid_caxis_spherical_harmonics(data, caxis_spherical_harmonics):
+    """Check if the spherical harmonic channel axes are ``ìnt`` and do
```
f-brinkmann, line 59: Typo should be 'i' not 'ì'

spharpy/classes/audio.py
```diff
@@ -55,14 +55,46 @@ def _atleast_3d_first_dimension(data):
...
+        Spherical harmonic data.
+    caxis_spherical_harmonics : tuple
+        Specifies which axis of data holds the spherical harmonic coefficients.
+        Negative indexing, i.e., interpreted relative to the end of the array.
+
+    Raises
+    ------
+    ValueError
+        Raised if the number of spherical harmonic channels does not match
+        (n_max + 1)^2 for an integer n_max.
```
f-brinkmann, line 74: This is checked below. Can you update this please?

spharpy/classes/audio.py
```diff
@@ -55,14 +55,46 @@ def _atleast_3d_first_dimension(data):
...
 
 
-def _assert_valid_number_of_sh_channels(shape, sh_axis):
+def _assert_valid_caxis_spherical_harmonics(data, caxis_spherical_harmonics):
+    """Check if the spherical harmonic channel axes are ``ìnt`` and do
+       not exceed the number of dimensions of the data.
+
+    Parameters
+    ----------
+    data : numpy.ndarray
```
f-brinkmann, line 64: I would suggest to directly pass data.ndim

spharpy/classes/audio.py
```diff
@@ -254,6 +289,122 @@ def basis_type(self, value):
         else:
             _SphericalHarmonicBase.basis_type.fset(self, value)
 
+    def reshape(self, newshape):
+        """
+        Return reshaped copy of the SphericalHarmonicAudio object. Axes
+        containing the spherical harmonics coefficients can not be reshaped.
```
f-brinkmann, line 295: Did not think of this before to be honest. A different solution would be to allow freely reshaping and return an Audio either always or if SH channels are affected by reshaping.

spharpy/classes/audio.py
```diff
@@ -254,6 +289,122 @@ def basis_type(self, value):
...
+        # check if there are axes with valid sh_channels
+        # and determine new caxis_spherical_harmonics
+        new_caxis_sh = []
+        for caxis in self._caxis_spherical_harmonics:
+            n_sh_channels = old_cshape[caxis]
+            if n_sh_channels not in new_cshape:
+                raise ValueError(
+                    "The requested new shape does not leave a "
+                    "channel axis to hold the spherical "
+                    "harmonic channels.")
```
f-brinkmann, line 328: Not sure if this always works. Could fail in case the size of a non-SH channel matches that of an SH channel.

spharpy/spherical.py
```diff
@@ -307,42 +307,46 @@ def renormalize(data, channel_convention, current_norm, target_norm, axis):
...
     else:
         orders, _ = acn_to_nm(acn)
 
-    # One (re)normalization factor is computed per Ambisonics channel. To make
-    # sure that the factors can be applied, new axes must be added. This is
-    # done by reshaping to the following shape
-    shape = [1] * data.ndim
-
-    shape[axis[0]] = sh_channels
+    # calculate one renormalization factor per sh channel
```
f-brinkmann, line 317: Is this required? All SH channels have the same order and hence the same normalization will be applied everywhere.

spharpy/spherical.py
```diff
@@ -307,42 +307,46 @@ def renormalize(data, channel_convention, current_norm, target_norm, axis):
...
+        factor *= n3d_to_maxn(acn)
+
+    data_renorm = data.copy()
+
+    # apply factor for each SH axis
+    for a in axis:
+        shape = [1] * data.ndim
+        shape[a] = sh_channels
+
+        data_renorm *= factor.reshape(shape)
```
f-brinkmann, line 349: see comment above - the same factor is applied all the time, which is correct.

f-brinkmann (changes requested):
Thanks for the quick implementation. Reshaping seems to get really annoying. Maybe something we can discuss in general.

spharpy/spherical.py
tluebeck: I thinks it's required yes. By SH channels I mean all channels within one SH axis, which of require individual normalization factors.

spharpy/spherical.py
tluebeck: I think this is correct here. Correct me if I'm wrong. Maybe there is a way to do it more elegant in one or two lines, but I think for readability it is ok to loop over axis.

